PR #4, opened 2018-03-19
(docs) Add faq entry about dealing with multiple data-testids

audiolion:
**What**: docs faq addition

**Why**: I read the faq, and even though the previous question addresses it I thought `queryByTestId` would handle this case, it doesn't so this faq addition clarifies the issue along with suggested ways to address it.

**How**:  N/A

**Checklist**:

* [x] Documentation
* [ ] Tests N/A
* [x] Ready to be merged 

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -18,6 +18,16 @@
         "infra",
         "test"
       ]
+    },
+    {
+      "login": "audiolion",
+      "name": "Ryan Castner",
+      "avatar_url": "https://avatars1.githubusercontent.com/u/2430381?v=4",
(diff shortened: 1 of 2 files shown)


.all-contributorsrc
```diff
@@ -1,6 +1,7 @@
 {
   "projectName": "react-testing-library",
   "projectOwner": "kentcdodds",
+  "repoType": "github",
```
audiolion: interesting, I ran `npm run add-contributor`, it put this in the wrong spot I guess?

kentcdodds: No worries. there was an update to `all-contributors` and I haven't updated my template yet :)

PR #10, opened 2018-03-21
feat: Add getByTestId utility

pbomb:
**What**:

This adds the `getByTestId` utility function.

**Why**:

To work better with typical use case of expecting queried elements to exist instead of potentially being null. 

**How**:

`queryByTestId` still exists and still can return null, but `getByTestId` is added that will throw an Error instead of returning null. For consumers using Flow or TypeScript, using this new function will not complain about directly using the returned element, since it cannot return `null`.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 


README.md
```diff
@@ -148,16 +148,27 @@ unmount()
...
+A shortcut to `` container.querySelector(`[data-testid="${yourId}"]`) ``. This could return null if no matching element is found. Read
 more about `data-testid`s below.
 
 ```javascript
-const usernameInputElement = queryByTestId('username-input')
+const hiddenItemElement = queryByTestId('item-hidden')
+expect(hiddenItemElement).toBeFalsy() // we just care it doesn't exist
+```
+
+#### `getByTestId`
```
kentcdodds: Let's move this above `queryByTestId` to indicate that it's recommended.

README.md
```diff
@@ -148,16 +148,27 @@ unmount()
...
+```javascript
+const usernameInputElement = getByTestId('username-input')
+usernameInputElement.value = 'new value'
+Simulate.change(usernameInputElement)
 ```
 
 ## More on `data-testid`s
 
-The `queryByTestId` utility is referring to the practice of using `data-testid`
+The `queryByTestId` and `getByTestId` utilities refer to the practice of using `data-testid`
```
kentcdodds: Swap the order here to give preferential treatment to `getByTestId`

README.md
```diff
@@ -286,7 +299,7 @@ Or you could include the index or an ID in your attribute:
 <li data-testid={`item-${item.id}`}>{item.text}</li>
 ```
 
-And then you could use the `queryByTestId`:
+And then you could use the `queryByTestId` utility:
```
kentcdodds: Could you switch this to `getByTestId`?

Also, could you add a FAQ for: `What if I want to verify that an element does NOT exist`? Where we show how to use `queryByTestId` as well as `container.querySelector` as possible solutions.

pbomb:
Thanks, @kentcdodds I'll implement your suggestions this afternoon.

src/index.js
```diff
@@ -7,16 +7,26 @@ function select(id) {
...
-function queryDivByTestId(div, id) {
+function queryByTestId(div, id) {
   return div.querySelector(select(id))
 }
 
+// we may expose this eventually
+function getByTestId(div, id) {
+  const el = queryByTestId(div, id)
+  if (!el) {
+    throw new Error(`Unable to find element by ${select(id)}`)
```
antoaravinth: @kentcdodds Oops! If now `getByTestId` throws error, then `toBeInTheDOM` (which I'm working on it) wouldn't make sense right?

kentcdodds: Yeah, I'd thought of that. It still works for `queryByTestId` though.

PR #13, opened 2018-03-22
Improving api

antoaravinth:
**What**: Improving test API

**Why**: Discussed already here : https://github.com/kentcdodds/react-testing-library/issues/3

**How**: Again as discussed :)

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

Related to this thread : https://github.com/kentcdodds/react-testing-library/issues/3


src/jest-extensions.js
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+      return {
+        message: () => 'expected the dom not to be present',
+        pass: false,
+      }
+    }
+  },
+
+  toHaveTextContent(htmlElement, checkWith) {
+    const textContent = htmlElement.textContent
```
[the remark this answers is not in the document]
antoaravinth: I guess the code that you had mentioned `getDisplayName` is enough. npm packages are there, but they offer many functionalities which we won't need at this time.

antoaravinth:
@kentcdodds Hey, have addressed your comments also I have moved our assertions to `element-queries`, I guess now that's the apt place. Also all the assertions are moved to jest-color, so we get nice color when things get failed :) Let me know if everything is good to go..

src/__tests__/element-queries.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 import {render} from '../'
+import '../../extend-expect'
```
kentcdodds: The problem with this is it means you have to run the build before you can run the tests (because on a fresh clone the `dist` directory doesn't work). It also means that any changes to `jest-extensions.js` wont be picked up by tests until the next build.

Instead, you'll basically need to duplicate the code.

Alternatively... You could put the code that's currently in `extend-expect.js` into `src/extend-expect.js` and then change the contents of: `extend-expect.js` to be simply: `require('./dist/extend-expect')`. Then you could update this import to simply: `'../extend-expect'` :+1:

src/__tests__/element-queries.js
```diff
@@ -66,4 +67,18 @@ test('totally empty label', () => {
   expect(() => getByLabelText('')).toThrowErrorMatchingSnapshot()
 })
 
+test('using jest helpers to assert element states', () => {
+  const {queryByTestId, getByTestId} = render(
+    <span data-testid="count-value">2</span>,
+  )
+
+  //other ways to assert your test cases, but you don't need all of them.
```
kentcdodds: Not sure why prettier didn't add a space here, but please add a space before the `o` :)

src/__tests__/element-queries.js
```diff
@@ -66,4 +67,18 @@ test('totally empty label', () => {
...
+  expect(queryByTestId('count-value')).toBeInTheDOM()
+  expect(queryByTestId('count-value')).toHaveTextContent('2')
+  expect(getByTestId('count-value')).toSatisfyDOM(el => el.textContent === '2')
```
kentcdodds: I'm starting to question the value of this matcher. Maybe instead we should just recommend people use [jest-extend](https://github.com/jest-community/jest-extended#tosatisfypredicate) instead. This is basically the same as `toSatisfy` from there.

The other matchers are still very valuable though!

antoaravinth: @kentcdodds `toSatisfyDOM` seems to be a fit for me in this library. Why we need to ask user to use jest-extend for a such a simple call? Yes, its almost the same code, but I feel it should be added. Anyways, let me know if you think otherwise.

kentcdodds: I agree that this matcher can be useful, but I'd much rather be very selective about what we include in this library. It's much easier to exclude something and add it later than add something and remove it later. So I lean on the side of excluding things. I'm thinking most people wont need this and those who do will be happy to include jest-extend I think.

src/jest-extensions.js
```diff
@@ -0,0 +1,102 @@
...
+  toHaveTextContent(htmlElement, checkWith) {
+    if (!(htmlElement instanceof HTMLElement))
+      throw new Error(
+        `The given subject is a ${getDisplayName(
+          htmlElement,
+        )}, not an HTMLElement`,
+      )
+
+    const textContent = htmlElement.textContent
+    const pass = textContent === checkWith
```
kentcdodds: Could we extend this to use the `matcher` function that's in the `queries.js` file? Perhaps we should move that `matcher` function to a `utils` file.

Using the `matcher` function would allow the `toHaveTextContent` to use a regex/case insensitive substring/function which I think would make this more powerful.

antoaravinth:
@kentcdodds Sure, will look into the changes and update the PR.

kentcdodds:
Also, let's figure out what what extra coverage we need for our tests. Open `coverage/lcov-report/index.html` in your browser to see the coverage report.

src/jest-extensions.js
antoaravinth: Yes, thats a good idea. Can take care of that, will move `matcher` to `utils` folder. Import `matcher` in `queries.js` and `jest-extension.js` files to use them. So fundamentally the call would be:

```
const pass = textContent === matches(textContent,htmlElement,checkWith)
```

antoaravinth:
@kentcdodds Yes, `jest-extensions.js` need to be tested. I can write few test cases for that.

antoaravinth:
@kentcdodds Update the PR. Few things that are pending:
1. Add test cases for `jest-extensions`
2. Write few more assertions for `toHaveTextContent` with `regex` etc, to show what user can actually do. 
3. Update the README on the following:

```
toHaveTextContent
toBeInTheDOM
```

their use cases etc. 

Anything else is missing?

src/utils.js
```diff
@@ -0,0 +1,9 @@
+export default function matches(textToMatch, node, matcher) {
```
kentcdodds: Rather than exporting this as the default (because this file is called `utils`), could we make it a named export (and feel free to disable the eslint warning, that's another one I've been meaning to disable).

```javascript
function matches() {/* etc... */}

export {matches}
```

antoaravinth:
@kentcdodds This should be good to go. Let me know if anything else is needed.

README.md
```diff
@@ -219,6 +219,31 @@ const usernameInputElement = getByTestId('username-input')
 > Learn more about `data-testid`s from the blog post
 > ["Making your UI tests resilient to change"][data-testid-blog-post]
 
+#### `toBeInTheDOM`
+
+There are two simple API which extend the `expect` API of jest for making assertions easier.
+The first one `toBeInTheDOM` which allows you to assert whether an element present in the DOM or not
+
+```javascript
+import 'extend-expect' //adds few API to jest's extend
```
kentcdodds: This will be `import 'react-testing-library/extend-expect'`

README.md
```diff
@@ -219,6 +219,31 @@ const usernameInputElement = getByTestId('username-input')
 > Learn more about `data-testid`s from the blog post
 > ["Making your UI tests resilient to change"][data-testid-blog-post]
 
+#### `toBeInTheDOM`
```
kentcdodds: Could we put these in a different section of the README (maybe right where it is).

So just change this to:

```md
## Custom Jest Matchers

There are two simple custom Jest matchers which you can use to extend the `expect` API of Jest for making assertions easier.

### `toBeInTheDOM`

This allows you to assert whether or not an element present in the DOM

... etc.
```

README.md
```diff
@@ -219,6 +219,31 @@ const usernameInputElement = getByTestId('username-input')
...
+expect(queryByTestId('count-value')).toBeInTheDOM()
+expect(queryByTestId('count-value1')).not.toBeInTheDOM()
+```
+
+#### `toHaveTextContent`
+
+This API allows you to check whether the given element has a text content or not.
+
+```javascript
+import 'extend-expect' //adds few API to jest's extend
```
kentcdodds: Here as well. Also, please add a space after the `//`

README.md
```diff
@@ -219,6 +219,31 @@ const usernameInputElement = getByTestId('username-input')
...
+
+render(<span data-testid="count-value">2</span>)
+expect(queryByTestId('count-value')).toHaveTextContent('2')
+expect(queryByTestId('count-value')).not.toHaveTextContent('21')
```
kentcdodds: For these two examples, could we use `getByTestId` rather than `queryByTestId`?

README.md
```diff
@@ -219,6 +219,31 @@ const usernameInputElement = getByTestId('username-input')
...
+
+There are two simple API which extend the `expect` API of jest for making assertions easier.
+The first one `toBeInTheDOM` which allows you to assert whether an element present in the DOM or not
+
+```javascript
+import 'extend-expect' //adds few API to jest's extend
+
+render(<span data-testid="count-value">2</span>)
+expect(queryByTestId('count-value')).toBeInTheDOM()
+expect(queryByTestId('count-value1')).not.toBeInTheDOM()
```
kentcdodds: Could you add a comment below that says: "Note: for this assertion, make sure you use a `query` function (like `queryByTestId`) rather than a `get` function (like `getByTestId`). Otherwise the get function could throw an error.

extend-expect.js
```diff
@@ -0,0 +1 @@
+require('./dist/extend-expect')
```
kentcdodds: Perfect

antoaravinth:
Sure Kent, will make these changes in sometime and update the PR.

antoaravinth:
@kentcdodds Addressed your comments. Let me know if anything else is needed?

kentcdodds:
Thanks so much for your help! I've added you as a collaborator on the project. Please make sure that you review the `other/MAINTAINING.md` and `CONTRIBUTING.md` files (specifically the bit about the commit messages and the git hooks) and familiarize yourself with the code of conduct (we're using [the contributor covenant](http://contributor-covenant.org/)). You might also want to [watch the repo](https://help.github.com/articles/watching-repositories/) to be notified when someone files an issue/PR. Please continue to make PRs as you feel the need (you can make your branches directly on the repo rather than your fork if you want). Thanks! And welcome to the team :)

PR #17, opened 2018-03-23
feat(render): add new query capabilities for improved tests

### LINKED ISSUE #16: Encourage using aria selectors
I've been in a discussion @neoziro on my blog where [he suggests](https://medium.com/@gregberge/yes-i-think-adding-a-aria-instead-of-a-data-test-is-much-more-valuable-for-the-user-c6e021b0be21) using `aria-` attributes and text content instead of `data-testid`. He makes great points and I think that it would be great if this library could provide selectors to encourage this behavior. I'm thinking that aria attributes wont always make sense for every test, in which case we could fallback to `data-testid`, and for [localization reasons](https://medium.com/@kentcdodds/im-glad-it-s-working-for-you-89210d06c438) text content may not make the most sense either, but making it easier to select based on text content could still make sense for some use cases and we could explain what those are.

So, out of this issue, I'd like to see some examples of what it would take to make utilities for selecting based on aria attributes as well as text content. If it works out, then we could recommend the following order of preference for selecting:

1. Text Content (99% of your users will be using your app by looking at the text content, so if you don't have issues with localization then this should be the best)
2. `aria-` attributes (your app should be accessible anyway, and this would encourage that as well).
3. `data-testid`s because they're way better than using class names, dom structure, or querying by React component classes.

Anyone want to try doing this?

kentcdodds:
**What**: Add the following methods

- queryByText
- getByText
- queryByPlaceholderText
- getByPlaceholderText
- queryByLabelText
- getByLabelText

**Why**: Closes #16 

These will really improve the usability of this module. These also align much better with the guiding principles :+1:

**How**:

- Created a `queries.js` file where we have all the logic for the queries and their associated getter functions
- Migrate tests where it makes sense
- Update docs considerably.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 
* [ ] Added myself to contributors table N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ components. It provides light utility functions on top of `react-dom` and
   * [`Simulate`](#simulate)
   * [`flushPromises`](#flushpromises)
   * [`render`](#render)
-* [More on `data-testid`s](#more-on-data-testids)
+* [`TextMatch`](#textmatch)
+* [`query` APIs](#query-apis)
 * [Examples](#examples)
 * [FAQ](#faq)
 * [Other Solutions](#other-solutions)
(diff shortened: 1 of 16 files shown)


kentcdodds:
I'm really excited about this!!

PR #19, opened 2018-03-24
[wip] feat(render) : add option for silencing render error

shrynx:
**What**:
added option to configure not throw/not throw error during render
solving the issue mention https://github.com/kentcdodds/react-testing-library/issues/18

**Why**:
To test custom error during render fails

**How**:
added try catch to ReactDOM 's render method and not throwing error if user specifies not to.

**Checklist**:

* [ ] Documentation
* [x] Tests
* [ ] Ready to be merged 

### DIFF
--- a/src/__tests__/throwRenderError.js
+++ b/src/__tests__/throwRenderError.js
@@ -1,0 +1,37 @@
+import React from 'react'
+import {render} from '../'
+
+global.console = {
+  error: jest.fn(),
+}
+
+const errorMessage = 'Please provide child component.'
+
+class ErrorComponent extends React.Component {
+  componentWillMount() {
+    if (!this.props.children) {
+      // eslint-disable-next-line
+      console.error(errorMessage)
+    }
+  }
+  render() {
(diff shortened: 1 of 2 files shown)


shrynx:
@kentcdodds  I am having issues with testing. although all the test pass, but they console.log error from react's side. Not sure how to solve it.
Also regarding documentation, the configure option of render (`container`) is not yet documented as `throwRenderError ` should be documented along side, i not sure how you would like it to be.

kentcdodds:
Hi. I'm sorry I don't have time for a proper response at the moment, but I should tell you before you do any more work on it that I won't be merging this pull request or supporting this feature. It goes against the guiding principles. I'll explain more later.

shrynx:
@kentcdodds alright, no worries 😃

shrynx:
@kentcdodds i realised i could just catch the render error in my test, thus also making my test better as it will check for error rather than trusting on silent failing.

Sorry for all the trouble with the issue and PR, i have closed the related issue too

PR #32, opened 2018-03-30
Message should be function and added test cases for the coverage as well

antoaravinth:
**What**: Jest `message` should be a function, we have returned a string, so in case of failure, the API is breaking. 

**Why**:

**How**: Moved them to functions and wrote coverage test cases for them as well. 

**Checklist**:

* [ ] Documentation
* [x] Tests
* [x] Ready to be merged 

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -95,6 +95,20 @@ test('using jest helpers to assert element states', () => {
   expect(() =>
     expect(queryByTestId('count-value2')).toHaveTextContent('2'),
   ).toThrowError()
+
+  // negative test cases wrapped in throwError assertions for coverage.
+  expect(() =>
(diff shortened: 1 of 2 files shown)


antoaravinth:
@kentcdodds Before I merge this, I wanted to check if the test cases I have added is mainly for coverage. Not sure `element-queries.js` is the right place to put. Yes, I have left the comment, but still want to cross check with you, if this is the right place to do so.

kentcdodds:
Remember that if you merge things yourself (generally I'd prefer to review PRs first) that you should use the Squash and Merge feature and that the commit message you set it to follows the convention. See `other/MAINTAINING.md` for more info :)

Thanks!

antoaravinth:
@kentcdodds Yes, sure. Will mostly give it for a review though. Thanks for merging it.

PR #41, opened 2018-04-03
Debug helper on `get` calls. 

antoaravinth:
**What**: This is already discussed here : https://github.com/kentcdodds/react-testing-library/issues/31

**Why**: Again discussed here : https://github.com/kentcdodds/react-testing-library/issues/31

**How**:

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,29 +71,30 @@ facilitate testing implementation details). Read more about this in
 ## Table of Contents
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
+
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
-
-- [Installation](#installation)
-- [Usage](#usage)
-  - [`render`](#render)
-  - [`Simulate`](#simulate)
-  - [`wait`](#wait)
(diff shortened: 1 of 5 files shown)


antoaravinth:
@kentcdodds I have just initiated the thread for printing the DOM element when `get` calls fails. I have added the code only for `getByText`. Here is how my console looks:

screen shot 2018-04-03 at 2 16 09 pm

Looks cool to me and definitely its going to help in debugging the process as discussed. 

Things that are pending:
1. We need to set some limit, may be we can expose that as a env variable something like, `DEBUG_PRINT_LIMIT`, else fallback to `7000` as discussed. 
2. Update all the `get` calls with the same logic. 
3. Update the readme about the point 1. 

I guess that should be fine, let me know what you feel. 

Note: The sample test case will be failing, anyways will fix once we are good to merge this PR.

antoaravinth:
Cool! Will get back on this by weekend.

antoaravinth:
@kentcdodds Updated the codebase, let me know if anything else is pending.

kentcdodds:
Could we move this over to [dom-testing-library](https://github.com/kentcdodds/dom-testing-library)?

Sorry to change things underneath you!

PR #44, opened 2018-04-05
Add toHaveAttribute custom matcher (closes #43)

### LINKED ISSUE #43: toHaveAttribute custom matcher
I'm finding my self commonly checking if an element has a given attribute or not, and possibly also expecting a certain value off some attribute.

```js
expect(getByTestId('ok-button').hasAttribute('disabled')).toBeTruthy()
expect(getByTestId('dropdown-toggle').getAttribute('aria-expanded')).toEqual('false')
```

I wonder if it would be acceptable to the goals and intentions of this library to provide certain other custom matchers for common tasks. In this particular case I'd be proposing a `.toHaveAttribute` custom matcher that could work both to assert that an element has an attribute present or not, but also to assert that the attribute is present and has a certain value:

```js
expect(getByTestId('ok-button')).toHaveAttribute('disabled')
expect(getByTestId('dropdown-toggle')).toHaveAttribute('aria-expanded', 'false')
```

It could be considered syntactic sugar, but I think it makes tests more readable. Although I'd also understand if these are considered out of the scope of this library.

gnapse:
Feedback is welcome on the code to build the message for the custom matcher. It got a bit tricky but I think it's worth it to give human-friendly messages of what went wrong, and there are quite a few cases to cover there.

Closes #43


src/jest-extensions.js
```diff
@@ -1,4 +1,12 @@
-import {matcherHint, printReceived, printExpected} from 'jest-matcher-utils' //eslint-disable-line import/no-extraneous-dependencies
+//eslint-disable-next-line import/no-extraneous-dependencies
+import {
+  matcherHint,
+  printReceived,
+  printExpected,
+  stringify,
+  RECEIVED_COLOR as receivedColor,
```
antoaravinth: Just wondering here, can't we use `printReceived`? Is that `RECEIVED_COLOR` does something apart from `printReceived`?

src/jest-extensions.js
```diff
@@ -75,6 +96,42 @@ const extensions = {
...
+        const expectedAttribute = isExpectedValuePresent
+          ? expectedColor(`${name}=${stringify(expectedValue)}`)
+          : expectedColor(name)
+        const expectedMsg = `Expected the element ${to} have attribute ${expectedAttribute}`
+        const receivedMsg = getAttrReceivedMsg(
+          hasAttribute,
+          expectedValue,
+          receivedValue,
+        )
+        const matcher = matcherHint(
```
antoaravinth: I have used `assertMessage` method to handle the messages. Can't we re-use them? Like I have done it over here: https://github.com/gnapse/react-testing-library/blob/bfeb26aa29e2b36cbf254a6cb8dc6934e32a055e/src/jest-extensions.js#L89

antoaravinth:
@gnapse Thanks for the wonderful PR! Left few comments!

src/jest-extensions.js
gnapse: @antoaravinth Not in this case. The messages I wanted to generate do not comply with the predefined format that `assertMessage` uses.

BTW, not even the two already existing custom matchers used this function. It's only being used in one of them. I do not think we can expect all or even most matchers to always adhere to this expected/received constrained message format. I was checking at jest's own codebase and how they approach generating messages, and the most common functionality they have across all them is encoded in all the functions they also publish in `jest-matcher-utils` and that we're using here.

src/jest-extensions.js
gnapse: Yes. `printReceived` uses `RECEIVED_COLOR` internally, but it also stringifies the argument. But in the new code I added I wanted to show `attribute-name="value"` all in the received or expected color, without stringifying all of it. If I used `printReceived`/`printExpected` I'd have the whole `attribute-name="value"` itself surrounded by double quotes (and possibly those inner quotes escaped then).

src/jest-extensions.js
antoaravinth: Yes, I got it. Just wanted to check. Cool, thanks!

gnapse:
Sure, here it is, the four different possible cases:

```
// Expect attribute to be present and with a certain value
.toHaveAttribute(attr, value)
.not.toHaveAttribute(attr, value)

// Expect attribute to be present, regardless of its value
.toHaveAttribute(attr)
.not.toHaveAttribute(attr)
```

image

kentcdodds:
Awesome. Could you update it so it resembles the errors from other assertions?

Like this:

screen shot 2018-04-05 at 1 10 14 pm

Specific differences are:

1. There's a `chalk.dim` comment after the assertion that could be handy: `// element.getAttribute('aria-expanded') === 'true'`. I'm fine to add `chalk` as a dependency for this :+1: Most people will already have it installed anyway.
2. Add a colon after the "Expected the element..." and the "Instead:"
3. Add a newline and two spaces before the expected and actual

Let me know if you have questions :)

gnapse:
@kentcdodds I did points 2 and 3 above.

However, regarding the first point, that is apparently a relatively new feature of the jest helper `matcherHint`, passing to it a `comment: '...'` in the last argument that's for options. So in theory I'd not even need to add a direct dependency to `chalk`, because `matcherHint` generates this internally.

However, it was odd that I did not get that comment when using `.toBe` in my own app's tests. That's when I realized I'm not on the latest jest version, and though I can upgrade, you can't guarantee that all the users of this lib will be either.

Do you prefer if I add the comment myself using `chalk` directly? Or would it be better to add it in this `comment` option, and it'll appear to users or not, depending on the version of jest that they're using.

kentcdodds:
Let's use the `comment` option. That's great to know! I wasn't aware of that feature :)

gnapse:
Me neither! I learned it while working on this. That's a huge part of what I love about contributing to open source.

src/jest-extensions.js
```diff
@@ -9,10 +17,30 @@ function getDisplayName(subject) {
...
+
+function getAttributeComment(name, value) {
+  return value === undefined
+    ? `element.hasAttribute(${stringify(name)})`
+    : `element.getAttribute(${stringify(name)}) === ${stringify(value)}`
+}
+
 const extensions = {
   toBeInTheDOM(received) {
     getDisplayName(received)
```
kentcdodds: I just noticed that this actually isn't doing anything. I'm pretty sure we should assign it a value called `displayName` and use that somehow or remove it entirely... Probably good for another PR though.

gnapse: Yes, I noticed that too, but I preferred no to touch it here as it's unrelated.

I do have a few code cleanup proposals for this file, that I was meaning to provide in a separate PR. For instance, the pattern of returning something different on the matchers depending on wether they `pass`ed or not seems redundant to me.

kentcdodds: Yes, I'd love it if you could clean this file up a bit in another PR! Let's just see what we need to do to increase coverage back to 100%.

Open `coverage/lcov-report/index.html` in your browser to see what's left :+1:

PR #46, opened 2018-04-06
Refactor custom matchers (closes #45)

### LINKED ISSUE #45: CleanUp for `toBeInTheDom` and `toHaveTextContent`
Following up the conversation on https://github.com/kentcdodds/react-testing-library/pull/44, looks like we need to do bit of clean up on `toBeInTheDom` and `toHaveTextContent` on the jest extensions. I can pick this up and finish this task.

@gnapse Let me know if you have already picked this up.

gnapse:
This is a proposed refactor to the custom matchers code, as briefly discussed in https://github.com/kentcdodds/react-testing-library/pull/44#discussion_r179593891:

- [x] Use common message format across all custom matchers
- [x] Make it easy to adopt the same format in any future matchers
- [x] Simplify code, remove unneeded if/else logic

Closes #45

#### Proposed changes not yet performed (pending discussion)

- [ ] Make each custom matcher a named export, instead of bundling them in an object as default export
- [ ] Better organize in a sub folder with one file per matchers and some utils file for common stuff

#### Pending to be fixed

- [x] Code coverage is not full because `getDisplayName` never gets called during the tests with a class as argument, only with html elements. I wonder if that is expected, given that this library is only dealing with rendered html and not the virtual dom, or is it?

### DIFF
--- a/src/__tests__/element-queries.js
+++ b/src/__tests__/element-queries.js
@@ -115,6 +115,10 @@ test('using jest helpers to assert element states', () => {
   expect(() =>
     expect(queryByTestId('count-value')).not.toHaveTextContent('2'),
   ).toThrowError()
+
+  expect(() =>
+    expect({thisIsNot: 'an html element'}).toBeInTheDOM(),
+  ).toThrowError()
 })
 
 test('using jest helpers to check element attributes', () => {
--- a/src/jest-extensions.js
+++ b/src/jest-extensions.js
@@ -52,6 +52,9 @@ function getAttributeComment(name, value) {
 
 const extensions = {
   toBeInTheDOM(received) {
+    if (received) {
+      checkHtmlElement(received)
+    }
     return {
       pass: !!received,
       message: () => {
--- a/src/jest-extensions.js
+++ b/src/jest-extensions.js
@@ -44,28 +44,15 @@ function getAttributeComment(name, value) {
 const extensions = {
   toBeInTheDOM(received) {
     getDisplayName(received)
-    if (received) {
-      return {
-        message: () =>
-          `${matcherHint(
-            '.not.toBeInTheDOM',
-            'received',
-            '',
-          )} Expected the element not to be present` +
-          `\nReceived : ${printReceived(received)}`,
-        pass: true,
-      }
-    } else {
-      return {
-        message: () =>
-          `${matcherHint(
-            '.toBeInTheDOM',
-            'received',
-            '',
-          )} Expected the element to be present` +
-          `\nReceived : ${printReceived(received)}`,
-        pass: false,
-      }
+    return {
+      message: () =>
+        `${matcherHint(
+          `${this.isNot ? '.not' : ''}.toBeInTheDOM`,
+          'received',
+          '',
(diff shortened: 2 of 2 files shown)


antoaravinth:
Thanks for this!

>  Better organize in a sub folder with one file per matchers and some utils file for common stuff

There are only 3 custom matchers, with few common utils like `getMessage`, `checkHtmlElement`. Do we really move them per file basis? May be / May be not :)

> Code coverage is not full because getDisplayName never gets called during the tests with a class as argument, only with html elements. I wonder if that is expected, given that this library is only dealing with rendered html and not the virtual dom, or is it?

What if someone does this:

```
expect(queryByText('notindom').toBeInTheDOM()
```

then we decided to throw exception like: 

```
The given subject is a object, not an HTMLElement
```

And also I guess you had to add `checkHtmlElement` for `toBeInTheDOM`.

gnapse:
Agree that it might not be a good idea to split the custom matchers. That's why I did not do it, just wanted to throw the idea out there. What do you think about exporting each of them individually as named exports? Not that's too important right now, but I find it odd to have them all bundled in an object, when they are independent pieces.

#### About this

> And also I guess you had to add checkHtmlElement for toBeInTheDOM.

The whole point of `.toBeInTheDOM` is that it could be called with something that is not found in the DOM, so it won't be a HTML element. We cannot throw an exception in that case. The actual matcher will take care of that when we return `pass: false`.

#### Regarding the `getDisplayName` issue

I guess what I need to do is to add a test case where we expect something that's not null (but also not a html element) to be in the DOM. That has to be a test that does not use `queryByText` or any of the other `query*` functions, because those all return either a html element of `null`. I'll do that.

gnapse:
Ok, sorry about the rushed response. I fully get what you mean now. I added the check for html elements in `toBeInTheDOM`, but only when the argument is not null. If the argument is null or undefined I leave it to fail normally, because it does not pass.

I was confused at first because this check was never there before, just a superfluous call to `getDisplayName` whose return value was not being used.

antoaravinth:
> Agree that it might not be a good idea to split the custom matchers. That's why I did not do it, just wanted to throw the idea out there. What do you think about exporting each of them individually as named exports? Not that's too important right now, but I find it odd to have them all bundled in an object, when they are independent pieces.

May be we can take this up. 

> The whole point of .toBeInTheDOM is that it could be called with something that is not found in the DOM, so it won't be a HTML element. We cannot throw an exception in that case. The actual matcher will take care of that when we return pass: false.

Yes, but since we have checks like:

```
checkHtmlElement(htmlElement)
const textContent = htmlElement.textContent
```

What if the `htmlElement` is not an `HTMLInstance`? We might end up in other issues. Yes, as you said we have to remove the error logic and pass the control to `matchers`, so that it ends up in fail test cases.

kentcdodds:
Could we move this over to [dom-testing-library](https://github.com/kentcdodds/dom-testing-library)?

Sorry to change things underneath you!

gnapse:
No problem, it makes sense. I'll re-create the PR over there.

PR #52, opened 2018-04-11
Add a definition for TextMatch

thchia:
**What**:
Update typings by adding a `TextMatch` type.

**Why**:
TypeScript consumers cannot use RegExp or functions in the query helpers, due to the argument being typed as string-only.

**How**:
Add a `TextMatch` type and replaced the `string` argument type for the various query helpers with it.

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged 


kentcdodds:
Oops. Could you rebase and resolve merge conflicts? Just did the one in `.all-contributorsrc` then run `npx kcd-scripts contributors generate` to fix the README one 👍

thchia:
Hope I did it right... I was trying to actually rebase it rather than push another commit on but I came a little unstuck. Thanks for this library!

thchia:
Sorry, tried again but it seems when I rebased the upstream changes it always thinks there is a conflict.

PR #68, opened 2018-05-02
feat(examples): Add context examples

alexkrolick:
**What**:

- Add React context examples
- WIP: Update docs/separate examples in src from actual src tests

**Why**:

**How**:

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged 


examples/README.md
```diff
@@ -0,0 +1,35 @@
...
+pop up with React. Check out the `__tests__` directory for the different
+examples.
+
+## Setup
+
+The examples have a unique jest/eslint set up so the test files will resemble
+how they might appear in your project. (You'll see in the tests that we can
+`import {render} from 'react-testing-library'`).
+
+Another thing that we do for the tests is we have a `setup-test-env.js` file in
```
alexkrolick: Could this be made a bit more obvious by putting it in `jest.setup.js` or something more similar to how Jest is typically configured?

examples/jest.config.js
```diff
@@ -0,0 +1,11 @@
+const jestConfig = require('kcd-scripts/jest')
+
+module.exports = Object.assign(jestConfig, {
+  rootDir: __dirname,
+  displayName: 'example',
+  setupTestFrameworkScriptFile: require.resolve('../other/setup-test-env'),
+  moduleNameMapper: {
+    // this is just here so our examples look like they would in a real project
+    'react-testing-library': require.resolve('../src'),
```
alexkrolick: 👍

examples/README.md
kentcdodds: I've never heard of `jest.setup.js` before. I'm not sure how much more obvious that is. Though I did consider just adding `import 'jest-dom/extend-expect'` in every file that needs the extensions to make it more obvious. That's probably the best idea. I'll do that.

alexkrolick: Actually it's fine as-is. I think maybe long-term the examples should drop the dependency on kcd-scripts to show a more direct Jest config, but that can be done later.

alexkrolick: nvm about `jest.setup.js` - it was manually referenced in my jest config, not automatic.

kentcdodds:
Whelp, I'm not sure what's up with codecov, but this is good to go!

kentcdodds:
Haha, whoops, this probably didn't need to be released 😅 Oh well.

examples/README.md
kentcdodds: I think that we need to have an example repo (somewhere else) that shows a good setup. Maybe one that uses create-react-app as well. We could just link to those in the examples.

PR #72, opened 2018-05-04
chore: make sub-APIs h3s

alexkrolick:
Kick the build script 😄 

**What**:

**Why**:

**How**:

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged 

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,13 @@ facilitate testing implementation details). Read more about this in
 * [Installation](#installation)
 * [Usage](#usage)
   * [`render`](#render)
+  * [`container`](#container)
+  * [`rerender`](#rerender)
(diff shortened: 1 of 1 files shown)


kentcdodds:
Whelp, looks like the codecov stuff is busted :( Something in this logic isn't happening properly:

https://github.com/kentcdodds/kcd-scripts/blob/ae04401a6cc0e6e02d2050e0af164098d7f3608f/src/scripts/travis-after-success.js#L11-L16

Either that or something here: 

https://github.com/kentcdodds/kcd-scripts/blob/ae04401a6cc0e6e02d2050e0af164098d7f3608f/src/scripts/travis-after-success.js#L33-L47

Pretty odd...

PR #74, opened 2018-05-04
fix: update getByTestId types

alexkrolick:
See https://github.com/kentcdodds/dom-testing-library/pull/29

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -2,13 +2,14 @@ import {Simulate as ReactSimulate} from 'react-dom/test-utils'
 
 type TextMatchFunction = (content: string, element: HTMLElement) => boolean
 type TextMatch = string | RegExp | TextMatchFunction
+type ExactTextMatch = string | RegExp | TextMatchFunction
 
 interface RenderResult {
   container: HTMLDivElement
   rerender: (ui: React.ReactElement<any>) => void
   unmount: VoidFunction
-  queryByTestId: (id: TextMatch) => HTMLElement | null
-  getByTestId: (id: TextMatch) => HTMLElement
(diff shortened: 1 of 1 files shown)


alexkrolick:
I'm actually wondering in getByAltText and other queries should also switch to exact match by default - since regex and functions are supported case-insensitive-partial-match seems a bit gratuitous.

kentcdodds:
> I'm actually wondering if getByAltText and other queries should also switch to exact match by default - since regex and functions are supported case-insensitive-partial-match seems a bit gratuitous.

That's fair. We should make that change in a major version bump though. I was thinking that if you wanted to do a more specific match you could use a regex with `^` and `$`, but maybe we should make it the inverse: use a regex to be more loose and a string to be more strict.

The thing is, if we consider it from a user's standpoint, they don't care whether the letters are capitals or wither the button says: "Submit form" or "Submit". It's all the same to them. I think that's what motivated the partial match for strings in my mind.

alexkrolick:
> if we consider it from a user's standpoint,

True, but there are other standpoints to consider, i.e., the product spec says button text should be capitalized or a styleguide says all copy should end in `.`, or for testing compliance with some user-provided theme/translation.

> but maybe we should make it the inverse: use a regex to be more loose and a string to be more strict.

I think this makes sense because the most basic "includes text" regex is just `/mysubstring/` which is almost the same as `"mysubstring"`.

I could see implementing this by exposing the final options argument to the `getAllByAttribute` to the public methods that use it so that that `exact: true` could toggled.

https://github.com/kentcdodds/dom-testing-library/blob/master/src/queries.js#L73

PR #77, opened 2018-05-06
Fix/examples types

alexkrolick:
**What**:

**Why**:

**How**:

**Checklist**:

* [ ] Documentation
* [ ] Tests
* [ ] Ready to be merged 


alexkrolick:
@maddijoyce the typescript definitions are almost certainly wrong 😃

typings/index.d.ts
```diff
@@ -2,22 +2,44 @@ import {Simulate as ReactSimulate} from 'react-dom/test-utils'
...
+    text: TextMatch,
+    options?: TextMatchOptions,
+  ) => HTMLElement
+  queryByLabelText: (
+    text: TextMatch,
+    options?: TextMatchOptions,
+  ) => HTMLElement | null
+  getByLabelText: (
+    id: TextMatch,
+    options?: {selector: string; TextMatchOptions},
```
maddijoyce: This should probably be
```options?: { selector?: string; } & TextMatchOptions```

alexkrolick: does that mean "all the keys from TextMatchOptions plus selector?"

maddijoyce: Yeah that's correct, it's a union type. I also made the selector optional, because it's * by default.

maddijoyce:
All looks good there now. As far as separating the types into their respective libraries, I'll put a PR through this week.

alexkrolick:
The inline type signatures for the wrapped methods don't have the new options yet, but maybe that can be a followup

PR #87, opened 2018-05-15
feat(debug): add a debug convenience method

### LINKED ISSUE #86: Add debug method

When I'm writing my tests I often find myself in a state when I need to know what's going on on the page. To accomplish this I usually write something like this: `getByText('something that for sure is not on the page')`. This way I have an error and the HTML is shown. I think it would be useful to expose a `debug` method that does more or less the same thing.

### Suggested implementation:

I guess we can reuse the same code used to show error messages

I tried to `console.log(container.innerHTML)` but it's all on the same line and hard to follow.

kentcdodds:
**What**: add a debug convenience method

Also add @Gpx to contributors!

**Why**:

Closes #86

**How**:

- Just re-export everything from dom-testing-library to simplify maintainence
- Add a simple `debug` method to make it really easy to log the container.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -256,6 +256,16 @@
         "bug",
         "code"
       ]
+    },
+    {
+      "login": "Gpx",
+      "name": "Giorgio Polvara",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/767959?v=4",
+      "profile": "https://twitter.com/Gpx",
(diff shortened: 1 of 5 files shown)


Gpx:
Hoped it helped. It's definitely going to be useful to me 🎉

Issue #62, opened 2018-04-20
How to test React.Portal


bugzpodder:
react-testing-library: 2.1.1
node: 8.9.3
yarn: 1.6.0

```
import React from "react";
import { render, Simulate } from "react-testing-library";

import Button from "material-ui/Button";
import Dialog, {
  DialogActions,
  DialogContent,
  DialogTitle
} from "material-ui/Dialog";

export const CommonDialog = props => {
  const { body, children, hideModal, isVisible, title } = props;
  return (
    <Dialog open={isVisible}>
      <DialogTitle>{title}</DialogTitle>
      <DialogContent>
        {body}
        {children}
      </DialogContent>
      <DialogActions>
        <Button id="close" onClick={hideModal}>
          Close
        </Button>
      </DialogActions>
    </Dialog>
  );
};

test("render dialog", () => {
  const mockCallback = jest.fn();
  const { getByText, getByTestId, container } = render(
    <CommonDialog title="test" isVisible={true} hideModal={mockCallback} />
  );
  Simulate.click(getByText("Close"));
  expect(mockCallback).toBeCalled();
  expect(container).toMatchSnapshot();
});
```
in the snapshot, it is just a simple div, and the Close button could not be found.  It is not immediately not obvious what's went wrong here.
I was using enzyme and it is working fine.

bugzpodder:
checked source code, Dialog uses Portal so maybe thats why?

kentcdodds:
Glad you found a workaround 👍

kentcdodds:
Make sure to unmount at the end of your test to clean up the document.body

ericmaicon:
Hey, this solve my problem while test a Component using reactstrap modal. I am wondering if there is a way to supress warnings, though.

![image](https://user-images.githubusercontent.com/1010637/39925887-d064a1d2-5503-11e8-8097-e9c3809b3f55.png)

bugzpodder:
`test("render dialog", () => {
	//  Disable Warning: render(): Rendering components directly into document.body is discouraged.
	const console = global.console;
	global.console = { error: jest.fn() };
	const { container, unmount } = render(
                 <Component />,
		{ container: document.body },
	);
	unmount();
	global.console = console;
});`

ericmaicon:
Thanks @bugzpodder. Worked!

kentcdodds:
[...]

PR #107, opened 2018-06-06
fix: add support for react portal elements

thesalah:
**What**:
Components like modals, dialogs, hovercards, tooltips etc use [React portals](https://reactjs.org/docs/portals.html) to render them into `document.body`. Testing utilities can query only on the container and this cannot do testing in such components.

**Why**:
This fix is necessary to test components that uses react portals.

**How**:
When using `renderIntoComponent`, instead of exposing test utilities for container, we do it  for `document.body` which contains the container.

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged 


kentcdodds:
This is very interesting! I'm leaning on accepting these changes but would like to hear what others have to say about it...

gnapse:
To add some context, #62 has a discussion about how this is possible today with some workaround (probably similar to what this PR brings out-of-the-box, haven't made a thorough comparison).

kentcdodds:
Perhaps we could expose some additional queries on a namespace:

```javascript
const {inBody} = renderIntoDocument(<Stuff />)
const submitButton = inBody.getByText('Submit')
```

kentcdodds:
But I'm not sure I like that either. I'm thinking the solution in #62 is very straightforward and prefer that over bloating the API further. Use cases for portals are few and far between so I don't think it's worth the extra overhead.

thesalah:
I'd prefer not to bloat the API for this. By documentation, `renderIntoDocument` says it render into document.body. So I hope just exposing test queries on document.body is meaningful here.

kentcdodds:
That makes sense to me. Ok, I'm in favor of this change. Let's make a change to the docs to explain this as well.

Issue #53, opened 2018-04-11
Variables in JSX strings put on new line


thchia:
* `react-testing-library` version: 2.1.0
* `node` version: 8.9.3
* `npm` (or `yarn`) version: `yarn` 1.3.2

Relevant code or config

Component to be tested:
```javascript
const Component = ({ currentStep }) => <div>Step {currentStep} of 4</div>
```

What you did:

Test code:
```javascript
const { getByText } = render(<Component currentStep={1} />)

expect(getByText('Step 1 of 4')).toBeDefined()
```

What happened:

```javascript
Unable to find an element with the text: Step 1 of 4. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

<div>
    Step
    1 // <--------- newline
      of 4
 </div>
```

Reproduction repository:

This can be reproduced in the current repository test cases by inserting a variable into the JSX.

Problem description:

Variables in JSX strings cause new line issues (and failing tests). I believe that in the spirit of "not testing implementation details", this should be considered a harmful problem. If I add an extra space around the "1", the test passes (period signifies space):

```javascript
expect(getByText('Step..1..of 4')).toBeDefined() // passing
```

Suggested solution:

I don't know why this is happening, but I'm guessing it is something to do with how the component is being serialised. I'm also guessing that this issue should be here and not dom-testing-library but I don't know for sure where the root of the problem lies.

gnapse:
Given that in html any extra whitespace between words in text is not meaningful, and the resulting text shown to the user on screen is what you expect in your test above, I think it makes sense to take this into account when matching an element's text content.

Actually I am now wondering if `.toHaveTextContent` should also take this into account.

kentcdodds:
Yes. We should update the `matcher` function in dom-testing-library to replace multiple spaces with one before it attempts to do any comparison. I consider this a bug and would love a pull request for it 👍

gnapse:
I'll take it, since I'll probably also have to do this to jest-dom's `.toHaveTextContent` ;)

DarrylD:
I'm still having this issue, any idea why it would continue to add white spaces? Updated to the latest dom-testing-library.

gnapse:
@DarrylD your best bet for others to be able to help is to provide a sample repository or codesandbox where the problem can be reproduced.

max-sr:
I have also continued to experience the same issue, and was able to diagnose it (as far as my case goes, at least).

[...]

PR #111, opened 2018-06-12
add within API

Dajust:
**What**: Add a `within` API

**Why**: Kindly see [issue #53 of dom-testing-library](https://github.com/kentcdodds/dom-testing-library/issues/53)

**How**: Exported `getQueriesForElement` as `within` from the index.js file

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 


src/index.js
```diff
@@ -44,4 +44,10 @@ syntheticEvents.forEach(eventName => {
 
 // just re-export everything from dom-testing-library
 export * from 'dom-testing-library'
```
kentcdodds: I was actually hoping this PR would be made to dom-testing-library. Then we'll get it automatically in react-testing-library because of this.

Dajust: Hi @kentcdodds, I noticed that dom-testing-library already exported getQueriesForElement as something else:

`export {getQueriesForElement as bindElementToQueries}`

Is it okay to still export it as yet another thing? like: 

`export {getQueriesForElement as bindElementToQueries}`
`export {getQueriesForElement as within}`

I've not done such export before :smile:

kentcdodds: Yup, that works just fine :+1: You can even do it in a single line:

```js
export {getQueriesForElement as bindElementToQueries, getQueriesForElement as within}
```

Dajust: Thanks man!

How about the docs, should I also move it to dom-testing-library?

kentcdodds: Might be good to have it in both places actually. I think we really need a documentation website to consolidate this, but I don't have time to build it right now 😅

Dajust: Have updated this to add only docs.

kentcdodds:
Could you resolve merge conflicts please? Thanks!

README.md
```diff
@@ -620,6 +621,22 @@ expect(submitButtons).toHaveLength(3) // expect 3 elements
 expect(submitButtons[0]).toBeInTheDOM()
 ```
 
+## `within` API
+
+Sometimes, there is no garauntee that the text, placeholder, or label you want to query is unique on the page. So you might want to explicity tell react-render-dom to get an element **only within** a particular section of the page. `within` is a helper function for this case.
```
klujanrosas: Small typo in here, `garauntee` should be `guarantee` :tada:

gnapse: What's react-render-dom?

PR #118, opened 2018-06-18
feat(render): deprecate renderIntoDocument and make it the default

### LINKED ISSUE #116: Deprecate renderIntoDocument()
The fundamental (and great) aspect of this library is:

> The more your tests resemble the way your software is used, the more confidence they can give you.

I've watched all your talks and you keep saying that `Simulate` is a bad idea (even Dan Abramov told us that, if I'm not wrong) and I really agree with you!

We all want to make testing easier, and the only confusing thing that I see in this great library is having two render methods.

My suggestion is that we just deprecate `renderIntoDocument()` and move `renderIntoDocument()`'s behavior to `render()`.

I don't see any reason to use `render()` over `renderIntoDocument()`.

kentcdodds:
**What**: deprecate renderIntoDocument and make it the default

**Why**: Closes #116 
 

**How**:

- Code refactor
- My son is waiting for me... Otherwise I'd be more informative.

**Checklist**:

* [x] Documentation
* [x] Tests
* [x] Ready to be merged 
* [ ] Added myself to contributors table N/A 

cc @hnordt @sompylasar

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,13 +12,11 @@
 
 [![Build Status][build-badge]][build]
 [![Code Coverage][coverage-badge]][coverage]
-[![version][version-badge]][package]
-[![downloads][downloads-badge]][npmtrends]
+[![version][version-badge]][package] [![downloads][downloads-badge]][npmtrends]
 [![MIT License][license-badge]][license]
 
 [![All Contributors](https://img.shields.io/badge/all_contributors-33-orange.svg?style=flat-square)](#contributors)
-[![PRs Welcome][prs-badge]][prs]
-[![Code of Conduct][coc-badge]][coc]
+[![PRs Welcome][prs-badge]][prs] [![Code of Conduct][coc-badge]][coc]
 [![Join the community on Spectrum][spectrum-badge]][spectrum]
 
 [![Watch on GitHub][github-watch-badge]][github-watch]
@@ -28,9 +26,9 @@
 ## The problem
 
 You want to write maintainable tests for your React components. As a part of
-this goal, you want your tests to avoid including implementation details of
-your components and rather focus on making your tests give you the confidence
-for which they are intended. As part of this, you want your testbase to be
+this goal, you want your tests to avoid including implementation details of your
+components and rather focus on making your tests give you the confidence for
+which they are intended. As part of this, you want your testbase to be
 maintainable in the long run so refactors of your components (changes to
 implementation but not functionality) don't break your tests and slow you and
 your team down.
@@ -39,10 +37,11 @@ your team down.
 
 The `react-testing-library` is a very light-weight solution for testing React
 components. It provides light utility functions on top of `react-dom` and
-`react-dom/test-utils`, in a way that encourages better testing practices.
-It's primary guiding principle is:
+`react-dom/test-utils`, in a way that encourages better testing practices. It's
+primary guiding principle is:
 
-> [The more your tests resemble the way your software is used, the more confidence they can give you.][guiding-principle]
+> [The more your tests resemble the way your software is used, the more
+> confidence they can give you.][guiding-principle]
 
 So rather than dealing with instances of rendered react components, your tests
 will work with actual DOM nodes. The utilities this library provides facilitate
@@ -66,8 +65,8 @@ facilitate testing implementation details). Read more about this in
 **What this library is not**:
 
 1.  A test runner or framework
-2.  Specific to a testing framework (though we recommend Jest as our
-    preference, the library works with any framework)
+2.  Specific to a testing framework (though we recommend Jest as our preference,
+    the library works with any framework)
 
 > NOTE: This library is built on top of
 > [`dom-testing-library`](https://github.com/kentcdodds/dom-testing-library)
@@ -81,13 +80,11 @@ facilitate testing implementation details). Read more about this in
 - [Installation](#installation)
 - [Usage](#usage)
   - [`render`](#render)
-  - [`renderIntoDocument`](#renderintodocument)
   - [`cleanup`](#cleanup)
-  - [`Simulate`](#simulate)
-  - [`wait`](#wait)
-  - [`waitForElement`](#waitforelement)
+- [`dom-testing-library` APIs](#dom-testing-library-apis)
   - [`fireEvent(node: HTMLElement, event: Event)`](#fireeventnode-htmlelement-event-event)
-  - [`prettyDOM`](#prettydom)
+  - [`waitForElement`](#waitforelement)
+  - [`wait`](#wait)
 - [`TextMatch`](#textmatch)
 - [`query` APIs](#query-apis)
 - [`queryAll` and `getAll` APIs](#queryall-and-getall-apis)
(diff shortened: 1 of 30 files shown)


hnordt:
The README needs to be updated. I can do it if you want.

kentcdodds:
What part? This PR makes major updates to the README.

hnordt:
@kentcdodds nevermind, it's there, I didn't see.

README.md
```diff
@@ -321,180 +366,61 @@ const usernameInputElement = getByTestId('username-input')
...
 instead of Synthetic Events. This aligns better with
 [the Guiding Principles](#guiding-principles).
+([Also Dan Abramov told me to stop use Simulate](https://twitter.com/dan_abramov/status/980807288444383232)).
 
-> NOTE: If you don't like having to render into the document to get `fireEvent`
-> working, then feel free to try to chip into making it possible for React
-> to attach event handlers to the rendered node rather than the `document`.
-> Learn more here:
+> NOTE: If you don't like having to use `cleanup` (which we have to do because
+> we render into document.body) to get `fireEvent` working, then feel free to
```
sompylasar: add backticks around `document.body`

README.md
```diff
@@ -526,32 +452,45 @@ fireEvent.click(getElementByText('Submit'), rightClick)
...
-console.log(prettyDOM(container))
-// <div>
-//   <h1>Hello World</h1>
-// </div>
+test('can fill in the form after loaded', async () => {
+  const {getByLabelText} = render(<Login />)
+
+  // wait until the callback does not throw an error. In this case, that means
+  // it'll wait until we can get a form control with a label that matches "username"
+  await wait(() => expect(queryByText(/loading.../i).not.toBeInTheDOM())
```
sompylasar: 1) _it'll wait until we can get a form control with a label that matches "username"_ does not sound aligned with the code it comments: it actually waits for "loading..." to disappear from the DOM, and only after that we look for the form control on the next line
2) `/loading.../i` – the `...` matches three arbitrary characters instead of the expected three ASCII dots, worth escaping?

src/__tests__/forms.js
```diff
@@ -32,29 +34,24 @@ function Login({onSubmit}) {
...
   // Act
   usernameNode.value = fakeUser.username
   passwordNode.value = fakeUser.password
   // NOTE: in jsdom, it's not possible to trigger a form submission
   // by clicking on the submit button. This is really unfortunate.
-  // So the next best thing is to simulate a submit on the form itself
+  // So the next best thing is to fire a submit on the form itself
   // then ensure that there's a submit button.
-  Simulate.submit(formNode)
+  fireEvent.click(submitButtonNode)
```
sompylasar: According to the `NOTE` comment above, clicking on the submit button won't work in `jsdom`. Does `fireEvent.click` take care of that?

src/index.js
```diff
@@ -2,39 +2,44 @@ import ReactDOM from 'react-dom'
...
-function cleanup() {
-  mountedContainers.forEach(container => {
-    document.body.removeChild(container)
-    ReactDOM.unmountComponentAtNode(container)
-    mountedContainers.delete(container)
-  })
+// maybe one day we'll expose this (perhaps even as a utility returned by render).
+// but let's wait until someone asks for it.
+function cleanupAtContainer(container) {
+  document.body.removeChild(container)
```
sompylasar: If `container` is custom passed by the caller, thus not appended to the `document.body` by `render`, the caller may not expect that it will be removed by `cleanup` which should be a mirror to `render`.

typings/index.d.ts
```diff
@@ -46,15 +46,14 @@ export interface RenderResult extends GetsAndQueries {
   debug: () => void
   rerender: (ui: React.ReactElement<any>) => void
   unmount: VoidFunction
+  cleanup: VoidFunction
```
sompylasar: Is this an established practice to have a single type definition for functions with different purposes? I would avoid sharing the type because these functions, although look similar, are not the same thing. For example, `debug` looks like a `VoidFunction` but the type is not shared. Oh by the way, I don't see a definition of `VoidFunction` in this file, where is it coming from?

src/index.js
kentcdodds: Hmmm... So what are you suggesting?

typings/index.d.ts
kentcdodds: This was a mistake on my part. Originally I had added a helper `cleanup` utility, then thought better of it because nobody's asking for that additional helper and not enough people would use it to be worth the docs.

src/index.js
kentcdodds: Oh, I see what you mean. I'm not sure how to solve that issue reliably. If they don't like the behavior then they can file an issue and we can add an option to prevent this. Let's wait to add complexity until someone has an issue.

sompylasar: Usually if a piece of code owns the lifecycle of an object (the caller in the case of custom container; the module with the render function in the case of default container), that piece is responsible for both creation+setup and teardown+destruction.

The suggestion is to make sure not to remove the container from the DOM if it hasn't been added to the DOM by the render module, only remove the map entry (lifecycle of which is owned by the render module) and `ReactDOM.unmountComponentAtNode` (which is a mirror to `ReactDOM.render`, mounted component's lifecycle is also owned by the render module).

I understand that the object/memory ownership and the allocate/deallocate pattern is often missed/violated in the JavaScript community due to various reasons, so it may look odd to require following it from this library users.

sompylasar:
I saw you removed the `cleanup: VoidFunction`, but the `unmount: VoidFunction` is still there in the `index.d.ts` TypeScript definition file, without a visible definition of the `VoidFunction` type – where does this type come from?

kentcdodds:
> where does this type come from?

No idea 🤷‍♂️

src/index.js
kentcdodds: That makes sense. Perhaps we should not add items to the set if we don't append it to the body. Would you like to open a PR for that?

sompylasar: We still need to keep a reference to unmount it, we could use a `Map<HTMLElement,boolean>` instead of `Set<HTMLElement>` to store whether the container was created (owned) or borrowed.

kentcdodds: Hmmmm..... I'm still not certain this is the behavior people would expect our want 🤔

kentcdodds: What if instead we did:

```javascript
container.parentElement && container.parentElement.removeChild(container)
```

I think that would make sense to happen in the cleanup regardless of whether it's in the body or not.

jomaxx:
@kentcdodds this is amazing! happy to see `renderIntoDocument` become the default `render`

antoaravinth:
Wow this is great @kentcdodds ! 👍

src/index.js
sompylasar: That would make sense to shorten the caller code to let it manage only `container` creation, and, my point still holds, the ownership of the `container` object lifecycle is moved from the caller into the `render` function.

kentcdodds: I'm a little confused by what you're suggesting. But why don't you make a PR and maybe that'll clear things up for me.

Issue #66, opened 2018-05-02
Simple text input value mutation. 


hally9k:
* `react-testing-library` version: 2.3.0
* `node` version: v8.9.1
* `npm` (or `yarn`) version: yarn@1.6.0

What you did:
I am just testing out this lib to see if I want to propose a switch from enzyme and I'm having trouble with the most basic stuff.

What happened:
Simply attempting to render an input, snapshot it, change the value and snapshot it again.

Reproduction repository: https://github.com/hally9k/react-testing-library-example

Problem description: Changing the value of an element doesn't reflect in the snapshot or invoke the onChange handler. 

Suggested solution: Explain to me what I'm missing as I clearly misunderstand something here.

alexkrolick:
You also need to simulate the change event:

```diff
  describe('Input > Text', () => {
	it('should render changes', () => {
		const mockOnChange = jest.fn()
		const { getByTestId } = render(
			<input type="text" value="hello" onChange={mockOnChange} data-testid="my-input" />
		)
		
		const input = getByTestId('my-input')

		expect(input).toMatchSnapshot()

		input.value = 'yo!'
+		Simulate.change(input)

		expect(input).toMatchSnapshot()
		expect(mockOnChange).toBeCalled()
	})
})
```

hally9k:
@alexkrolick Thanks for your answer. Simulating the change event will get the onChange handler to fire but the snapshot still doesn't reflect the change in the input's value attr. In enzyme you have to call component.update() to flush the changes. Is there something I'm missing here?

PR #120, opened 2018-06-20
Add note about triggering onChange with a different value

silvenon:
**What**: I added an explanation and an example how to trigger the `onChange` handler with `fireEvent`:

```js
const comment = getByLabelText('Comment')
comment.value = 'Great advice, I love your posts!'
fireEvent.change(comment)
```

**Why**: I didn't know how to do this because I was so used to `Simulate`, it took me a while to figure it out. I'm not 100% sure if this is the "right" way to do it, someone please correct me if it's not.

**Checklist**:

- [x] Documentation
- [ ] Tests (N/A)
- [x] Ready to be merged
      
This PR references #66.


kelly-tock:
found this after trying to figure out a form test, I noticed when you fire change, only the `target` property of event is fired and not `currentTarget`, 

example:
```
<input onChange={this.handleChange} value={value} data-testid="input" />
```

const input = getByTestId('input');

 input.value = 'testing';

 fireEvent.change(
    input
 );
```

in the change handler, target is defined but not current target. 

any ideas?

**Comment by kelly-tock** [comment]
also as an example, if you're using flowtype or typescript, 

```
handleChange(event: React.SyntheticEvent<HTMLInputElement>) {
    const value = event.currentTarget.value;
```

works, but target does not.

kentcdodds:
That could be a bug with dom-testing-library. Care to file that?

kelly-tock:
cool, i'll file it.

kelly-tock:
https://github.com/kentcdodds/dom-testing-library/issues/73

PR #130, opened 2018-07-02
update typings with dom-testing-library queries and export them

jpavon:
**What**: Update queries typings to match dom-testing-library

**Why**: Not up to date typings with dom-testing-library

**How**: Modify typings/index.d.ts.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged
      

jpavon:
@pbomb I changed the order to match with the order here: https://github.com/kentcdodds/dom-testing-library/blob/master/typings/queries.d.ts and they are exported because this functions can also be imported from 'react-testing-library'

pbomb:
Okay, making the orders consistent seems reasonable to do. :+1:

I still didn't think the dom-testing-library functions were re-exported from react-testing-library. Where do you see that in the code? @kentcdodds can you confirm if the dom-testing-library query functions are re-exported from react-testing-library?

kentcdodds:
Indeed it is: https://github.com/kentcdodds/react-testing-library/blob/de2c5cff55d336150cb9e973ccac38ae0d021c01/src/index.js#L57

jpavon:
@pbomb @kentcdodds I'm thinking having `export * from 'dom-testing-library'` inside the `index.d.ts` makes more sense and would remove a lot of duplication, let me know what you think I can create a pr in the next days.

kentcdodds:
Sounds good to me

PR #151, opened 2018-08-01
fix: Add baseElement to render return type

gnapse:
**What**:

Add `baseElement` to `render` return type.

**Why**:

Because without it TypeScript code using this library won't be able to use `baseElement`, which is crucial to testing react portals, for example.

**How**:

By adding the `baseElement` property to the type definitions.

**Checklist**:

- [x] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -6,6 +6,7 @@ type GetsAndQueries = ReturnType<typeof getQueriesForElement>
 
 export interface RenderResult extends GetsAndQueries {
   container: HTMLDivElement
+  baseElement: HTMLDivElement
   debug: (baseElement?: HTMLElement) => void
   rerender: (ui: React.ReactElement<any>) => void
   unmount: () => boolean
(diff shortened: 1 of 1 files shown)


gnapse:
Actually I now think there's an inconsistency in the type definitions. The `render` function is documented to optionally receive a `container: HTMLElement`, but it returns this very container (or a newly created one if none is provided), and the one returned is documented to be of type `HTMLDivElement`. What if the container that was passed in is not a `div`?

The above concern also applies to the `baseElement`. It is documented as an optional option of type `HTMLElement`, so the type of the returned one should also be `HTMLElement`, not `HTMLDivElement`.

pbomb:
I agree that `baseElement` and `container` should both be typed as `HTMLElement` instead of `HTMLDivElement`. I also wouldn't expect this change to impact hardly anyone as it seems rare that one would need to access div-specific properties. I'm not even sure what is div-specific, if anything at all.

gnapse:
Any ideas on when this can be merged? I wonder why the codecov task above does not finish.

kentcdodds:
I've been having trouble with codecov's status report for a month or two. I'm not sure what the problem is and haven't had time to investigate :-/

gnapse:
No problem, good to know. Thanks!

Issue #152, opened 2018-08-03
Form Inputs not updating after change event is fired


pcfields:
- `react-testing-library` version: 4.1.5
- `react` version: 16.4.1
- `node` version: 8.6.0
- `npm` (or `yarn`) version: 6.2.0

### Relevant code or config:

```js
test('submits form successfully', async () => {

const { getByText, queryByText, getByLabelText, debug } = renderWithRouter(
<SignIn setCurrentUser={jest.fn()} isSignedIn={false} history={{}} />
);

const submitButton = getByText('Sign In');
const emailInput = getByLabelText('Email');
const passwordInput = getByLabelText('Password');

emailInput.value = 'test@test.com';
passwordInput.value = '12345678';

fireEvent.change(emailInput);
fireEvent.change(passwordInput);

debug();
/* 
In debug email has value of 'test@test.com' and password has a value of ' ' 
*/

fireEvent.click(submitButton);

const emailErrorNode = queryByText('Email is required');
const passwordErrorNode = queryByText('Password is required')

expect(emailErrorNode).toBeNull();
expect(passwordErrorNode).toBeNull();

});
```

### What you did:

Test a sign in form by setting username to 'test@test.com' and password to '12345678'

### What happened:

When updating 2 inputs in a test, only one input gets updated the second input's value is ' ' (empty) 

### Reproduction:

https://codesandbox.io/s/0xwn98wo1p

### Problem description:

I can't test forms with more than one field change

kentcdodds:
Hi @pcfields,

Thanks for this. Please ping in here when you've added the *minimal* reproduction codesandbox. I definitely want to improve this so testing this kind of experience is more obvious.

FWIW, `change` is one of the events that we do some trickery for because the `change` event works differently with React than it does in a real browser. I'd love to do away entirely with this code: https://github.com/kentcdodds/react-testing-library/blob/9667967a085e438919fe3d02bc36761735bb2732/src/index.js#L50-L56

pcfields:
I've updated the codesandbox https://codesandbox.io/s/0xwn98wo1p

pcfields:
Let me know how I can help you fix this?

kentcdodds:
Sounds great! Could you strip down your example to be as minimal as possible? The less code in there to reproduce the issue the better as it makes it easier for us to determine what could cause the issue.

pcfields:
@kentcdodds I've stripped down the code. Let me know if this is enough.

[...]

PR #153, opened 2018-08-04
WIP: add setNativeValue for change event

kentcdodds:
**What**: This overrides the `fireEvent.change` function to one that allows users to set the target.value using a special `setNativeValue` function.

**Why**: (#152) Because React tracks when you set the `value` property on an input to keep track of the node's value. When you dispatch a `change` event, [it checks it's last `value` against the current value](https://github.com/facebook/react/blob/dd5fad29616f706f484938663e93aaadd2a5e594/packages/react-dom/src/client/inputValueTracking.js#L129) and if they're the same it does not call any event handlers (as no change has taken place as far as react is concerned). So we have to set the value in a way that [React's value setter function](https://github.com/facebook/react/blob/dd5fad29616f706f484938663e93aaadd2a5e594/packages/react-dom/src/client/inputValueTracking.js#L78-L81) will not be called, which is where the `setNativeValue` comes into play.

**How**: Development of the `setNativeValue` function was a team effort: https://github.com/facebook/react/issues/10135#issuecomment-401496776

This is kinda cheating because it's overriding `fireEvent.change`. I think I'd rather we do this within `dom-testing-library` directly.

I'm also not super happy with the API:

```javascript
fireEvent.change(input, {target: {value: 'hi'}})
```

It makes sense to me because in your `onChange` handler if you want the value you'll get it from `event.target.value`, but I worry that this API will make people think they can change other things about the `target` as well... I suppose we could make that work by Object.assigning all `target` properties onto the given `node` (while still treating the `value` property specially) 🤔 

Another thing I've considered is just accepting a `value` option:

```javascript
fireEvent.change(input, {value: 'hi'})
```

But I don't like that because the 2nd argument is _supposed_ to be event init options... I'm open to feedback here.

Also, this PR is incomplete because we still have to handle `select`, `mouseEnter`, and `mouseLeave`. I'm not sure how to handle those. More investigation is needed. 

I'm thinking that we should make these changes in `dom-testing-library` instead of here and we can address each of these one-by-one.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged
      
### DIFF
--- a/src/__tests__/events.js
+++ b/src/__tests__/events.js
@@ -30,7 +30,7 @@ const eventTypes = [
   },
   {
     type: 'Focus',
-    events: ['change', 'input', 'invalid'],
+    events: ['input', 'invalid'],
     elementType: 'input',
   },
   {
@@ -154,3 +154,14 @@ eventTypes.forEach(({type, events, elementType, init}) => {
     })
   })
 })
+
+test('onChange works', () => {
+  const spy = jest.fn()
+
+  const {container} = render(<input onChange={spy} />)
+  const input = container.firstChild
+  fireEvent.change(input, {target: {value: 'c'}})
+  expect(spy).toHaveBeenCalledTimes(1)
+})
+
+/* eslint complexity:0 */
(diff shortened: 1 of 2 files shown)


kentcdodds:
This is superseded by https://github.com/kentcdodds/dom-testing-library/pull/85

PR #166, opened 2018-09-03
Docs: add global setup, custom render docs

alexkrolick:


README.md
```diff
@@ -73,12 +73,54 @@ facilitate testing implementation details). Read more about this in
...
+test('Fetch makes an API call and displays the greeting when load-greeting is clicked', async () => {
+  // Arrange
+  axiosMock.get.mockResolvedValueOnce({data: {greeting: 'hello there'}})
+  const url = '/greeting'
+  const {getByText, getByTestId, container} = render(<Fetch url={url} />)
+
+  // Act
+  fireEvent.click(getByText('Load Greeting'))
+
+  // let's wait for our mocked `get` request promise to resolve
```
kentcdodds: Let's rephrase these comments:

```
// Let's wait until our mocked `get` request promise resolves and
// the component calls setState and re-renders.
// getByTestId throws an error if it cannot find an element with the given ID
// and waitForElement will wait until the callback doesn't throw an error
```

README.md
```diff
@@ -118,45 +160,99 @@ This library has a `peerDependencies` listing for `react-dom`.
 You may also be interested in installing `jest-dom` so you can use
 [the custom jest matchers](https://github.com/gnapse/jest-dom#readme)
 
-## Usage
+## Setup
```
kentcdodds: Let's add some text between `## Setup` and `### Globals` that says something like:

```
react-testing-library does not require any configuration to be used (as demonstrated in the example above). However, there are some things you can do to when configuring your testing framework to reduce some boilerplate. In these docs we'll demonstrate configuring Jest, but you should be able to do similar things with any testing framework (react-testing-library does not require that you use Jest).
```

README.md
```diff
@@ -118,45 +160,99 @@ This library has a `peerDependencies` listing for `react-dom`.
...
+
+There are several options you can add to your global test config that simplify
+the setup and teardown of tests in individual files. For example, you can ensure
+[`cleanup`](#cleanup) is called after each test and import additional
+assertions.
+
+To do this with Jest, you can add the
+[`setupTestFrameworkScriptFile`](https://facebook.github.io/jest/docs/en/configuration.html#setuptestframeworkscriptfile-string)
+option to your Jest config. The setup file can be anywhere, but if you're using
+create-react-app, use the default `src/setupTests.js`.
```
kentcdodds: Change to:

```
... use the default `src/setupTests.js` (in this case, no configuration changes are necessary).
```

README.md
```diff
@@ -118,45 +160,99 @@ This library has a `peerDependencies` listing for `react-dom`.
...
+- import { render, fireEvent } from '../test-utils';
++ import { render, fireEvent } from 'test-utils';
+```
+
+```diff
+// jest.config.js
+module.exports = {
+  moduleDirectories: [
+    'node_modules',
++    __dirname, // the project root
```
kentcdodds: Let's make the comment say: "the directory where your `test-utils.js` file resides."

I'd prefer to recommend that file exists in a directory called `test` rather than at the root of the project.

README.md
```diff
@@ -118,45 +160,99 @@ This library has a `peerDependencies` listing for `react-dom`.
...
-// this adds custom jest matchers from jest-dom
+// jest.config.js
+module.exports = {
+  setupTestFrameworkScriptFile: require.resolve('./jest.setup.js'),
+  // ... other options ...
+}
+```
+
+```javascript
+// jest.setup.js
```
kentcdodds: Maybe we should recommend putting this file in a directory called `test` to avoid overfilling the root directory?

alexkrolick: TBH I prefer putting that stuff in root initially and cleaning things up only if it gets overwhelming. Plus, `test` sounds like a place to put tests and not just a config directory.

README.md
alexkrolick: Not sure I understand "no configuration changes are necessary". Can you clarify?

alexkrolick: Oh I see. It doesn't require changing the Jest config. I think it can be worded better than that even:

>>>
To do this with Jest, you can add the
`setupTestFrameworkScriptFile` option to your Jest config.

> Note:
> If you are using the default setup from `create-react-app`, this option is set to `src/setupTests.js`. You should create this file if it doesn't exist and put the setup code there.

README.md
alexkrolick: Added a line here: https://github.com/kentcdodds/react-testing-library/pull/166/files#diff-04c6e90faac2675aa89e2176d2eec7d8R179

alexkrolick:
🎉

PR #177, opened 2018-09-18
New: expose option to override bound queries

alexkrolick:
**What**:

Useful if you want to add or change the queries

**Why**:

The custom render method documented in the README works well for wrapping the rendered node, but needs an extension point for adding/overriding queries. For example this would let you override the getByTestId query with a query for a different data-attribute (common request). 

**How**:

getQueriesForElement already takes a query argument, this exposes it in the render argument

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

alexkrolick:
I'll make a PR to document after using it more - might be part of a larger effort to show how to use custom queries

PR #192, opened 2018-10-05
feat: add `asFragment` return value from `render`

SimenB:
**What**:
Adds the possibility to get a `DocumentFragment` version of the rendered component.

**Why**:
I have a test of a connected redux component, and I want to snapshot the diff between them after dispatching an action. Since `container` is a live binding, its `innerHTML` changes, meaning comparison becomes meaningless.

An extra advantage is IMO that you don't need to do `firstChild` on snapshots, as having the `Fragment` wrapper makes sense. At least in my mind 🙂 

**How**:
By upgrading `kcd-scripts` to a version with Jest 23, and adding a dependency on `JSDOM` which matches the one found in Jest (to not rely on dependency hoisting).

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      
### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 import ReactDOM from 'react-dom'
 import {Simulate} from 'react-dom/test-utils'
 import {getQueriesForElement, prettyDOM} from 'dom-testing-library'
+import {JSDOM} from 'jsdom'
 
 const mountedContainers = new Set()
 
@@ -29,6 +30,7 @@ function render(ui, {container, baseElement = container, queries} = {}) {
       // Intentionally do not return anything to avoid unnecessarily complicating the API.
       // folks can use all the same utilities we return in the first place that are bound to the container
     },
+    asFragment: () => JSDOM.fragment(container.innerHTML),
     ...getQueriesForElement(baseElement, queries),
   }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,6 @@
 import ReactDOM from 'react-dom'
 import {Simulate} from 'react-dom/test-utils'
 import {getQueriesForElement, prettyDOM} from 'dom-testing-library'
-import {JSDOM} from 'jsdom'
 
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -29,6 +30,7 @@ function render(ui, {container, baseElement = container, queries} = {}) {
       // Intentionally do not return anything to avoid unnecessarily complicating the API.
       // folks can use all the same utilities we return in the first place that are bound to the container
     },
+    asFragment: () => JSDOM.fragment(container.innerHTML),
```
kentcdodds: Is there a reason to do this with JSDOM rather than:

```javascript
const fragment = document.createDocumentFragment()
fragment.innerHTML = container.innerHTML
```

That way we don't have to rely on JSDOM directly. I want this package to be capable of running in the browser.

SimenB: Great point! Pushed a new commit using plain DOM APIs for it.

Followed https://stackoverflow.com/a/25214113/1850276, but JSDOM doesn't support Ranges (https://github.com/jsdom/jsdom/issues/317), so I added a fallback for it

kentcdodds:
Would you like to [add yourself](../blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) to the [contributors table](../#contributors)?

src/index.js
```diff
@@ -29,6 +29,17 @@ function render(ui, {container, baseElement = container, queries} = {}) {
       // Intentionally do not return anything to avoid unnecessarily complicating the API.
       // folks can use all the same utilities we return in the first place that are bound to the container
     },
+    asFragment: () => {
+      if (typeof document.createRange === 'function') {
```
SimenB: might wanna add a `/* istanbul ignore if */` here, and leave a comment pointing back to https://github.com/jsdom/jsdom/issues/317

kentcdodds: Thanks! I've just committed that 👍

PR #207, opened 2018-10-23
docs: ✏️ explain how to implement custo getByTestId

Gpx:
**What**:

Adding an explanation on how to implement a `getByTestId` that looks for `data-test-id`.

**Why**:

Because people tend to ask this question.

**How**:

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [x] Ready to be merged
      
### DIFF
--- a/README.md
+++ b/README.md
@@ -557,6 +557,47 @@ const usernameInputElement = getByTestId('username-input')
 > about `data-testid`s from the blog post ["Making your UI tests resilient to
 > change"][data-testid-blog-post]
 
+<details>
+  <summary>What if my project already uses <code>data-test-id</code>? Do I have to migrate to <code>data-testid</code>?
+</summary>
+
+`getByTestId` is looking for the `data-testid` attribute and that's not going to
+change
+[#76](https://github.com/kentcdodds/dom-testing-library/issues/76#issuecomment-406321122)
+[#128](https://github.com/kentcdodds/dom-testing-library/issues/128)
+[#204](https://github.com/kentcdodds/react-testing-library/issues/204).
+
+What you can do is to create a [custom render](#custom-render) that overwrites
+`getByTestId`:
+
+```js
+// test-utils.js
+import {render} from 'react-testing-library'
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -557,6 +557,47 @@ const usernameInputElement = getByTestId('username-input')
...
+// test-utils.js
+import {render} from 'react-testing-library'
+
+const customRender = (node, ...options) => {
+  const utils = render(node, ...options)
+
+  return {
+    ...utils,
+    getByTestId: id => {
+      const el = utils.container.querySelector(`[data-test-id="${id}"]`)
```
alexkrolick: I think it would be good to use the query helpers from `dom-testing-library` to allow the custom function to support functions, strings, and regex and the `exact` and `preserveWhitespace` options in the same way as the other queries:

https://github.com/kentcdodds/dom-testing-library/blob/master/src/query-helpers.js#L36-L47

alexkrolick: https://github.com/kentcdodds/dom-testing-library/blob/master/src/queries.js#L153-L154

https://github.com/kentcdodds/dom-testing-library/blob/master/src/queries.js#L181-L190

https://github.com/kentcdodds/dom-testing-library/blob/master/src/queries.js#L192-L194

kentcdodds: True, feel fee to make another PR to improve that :) Thanks!

Gpx: I used the query helpers in a firs implementation but I was worried the example would go out of sync if the internals for `dom-testing-library` changed

kentcdodds: Good point. If it's an explicitly exported value, then it should be pretty solid. But if you're importing things from `dist` then that's probably not safe.

alexkrolick:
Query helpers are exported directly for this very use case
https://github.com/kentcdodds/dom-testing-library/blob/master/src/index.js--


- Alex

README.md
alexkrolick: Query helpers are exported directly for this very use case
https://github.com/kentcdodds/dom-testing-library/blob/master/src/index.js

Issue #206, opened 2018-10-23
Drop dependency on React's Simulate module


kentcdodds:
This code has got to go:

https://github.com/kentcdodds/react-testing-library/blob/fe3b65ca158f0ac088b67b195b678a9428a7a2b5/src/index.js#L62-L68

I'm not sure what is needed, but I suggest that we methodically remove one element from that array, see what breaks, then add some code until it starts working.

It may mean that we continue to monkey-patch `fireEvent` to support React events properly (because React is a little special). All I really care about is getting rid of the Simulate module. The biggest reason is #129, but also Dan told me we shouldn't use it and it's hacky and I want to get rid of it :)

weyert:
When you say doesn't work in the DOM you are referring to `jsdom` correct? As far as I know the mouse enter and mouse leave events exists as they are mentioned on MDN. The JSDOM does define the `MouseEvent` at: https://github.com/jsdom/jsdom/blob/13b2326b57ec621c4b75a64966ea23d89d8fce99/lib/jsdom/living/events/MouseEvent-impl.js

kentcdodds:
I think that comment is referring to the fact that they don't work in the same way the DOM works.

If you remove them from the array, you'll notice that tests start to break. Try to fix the tests without changing the tests. That's the task at hand :)

weyert:
I have been looking into this for a bit I find potentially relevant or interesting unit tests in the React repository: 

https://github.com/facebook/react/blob/b87aabdfe1b7461e7331abb3601d9e6bb27544bc/packages/react-dom/src/events/__tests__/EnterLeaveEventPlugin-test.js
and 
https://github.com/facebook/react/blob/b87aabdfe1b7461e7331abb3601d9e6bb27544bc/packages/react-dom/src/events/__tests__/SelectEventPlugin-test.js

kentcdodds:
@sivkoff fixed this! Thank you Vitaly!

sivkoff:
@kentcdodds you're always welcome!

Issue #129, opened 2018-06-29
Ship UMD build


jalvarado91:
### Describe the feature you'd like:

It would be awesome if there was a umd build so that this could be used with something like QUnit which is run in the browser. This way the lib could be included via a script tag either downloaded or with unpkg and be ready to use in environments without module loading.  

This might make more sense over in the dom-testing-library repo, but I'm putting it here since the intended use is with react. 

### Suggested implementation:

Similar to React or other libs that export umd builds, it would mean exporting exporting a global like `ReactTestingLib` or maybe something shorter (open to suggestions).

P.S. I can work on a PR for this feature. Seems like it would be fun. 

### Describe alternatives you've considered:

Switch test runners (big effort).

### Teachability, Documentation, Adoption, Migration Strategy:

Build example of using this in an environment with no module loading. Purely script tags in a page.

kentcdodds:
We can definitely do this. `kcd-scripts build` supports this via a `--bundle` flag. I think that'd probably be fine because people should be requiring the commonjs files directly anyway. Let me give this a shot...

kentcdodds:
Well, there's a UMD build for dom-testing-library now, but I'm struggling to get `react-dom/test-utils` to build with the UMD bundle of react-testing-library. I'd like to drop that dependency anyway though so I'm looking at ways to do that. I'm feeling pretty good about [my progress](https://github.com/kentcdodds/react-testing-library/blob/pr/stuffs-wip/src/index.js#L55-L69), but there are still some complications and things that need to be worked out thanks to how react does things with events. Hopefully when we can drop the Simulate dependency entirely though we'll be able to have a good UMD build :)

PR #210, opened 2018-10-24
Remove `Simulate` from dependencies

sivkoff:
**What**:
Removed `react-tom/test-utils` as a dependency

**Why**:
The biggest reason is #129 (see more details #206)

**How**:
By creating the `createReactEventSimulator` function that makes possible to simulate React events

**Checklist**:

- [x] Documentation N/A
- [x] Tests N/A
- [x] Ready to be merged
      

sivkoff:
@kentcdodds I've made rebase, but I'm concerned if I have broken something 😅

src/__tests__/events.js
```diff
@@ -156,10 +156,12 @@ eventTypes.forEach(({type, events, elementType, init}) => {
...
-  expect(handleChange).toHaveBeenCalledTimes(1)
+  const spy = jest.fn()
+
+  const {container} = render(<input onChange={spy} />)
+  const input = container.firstChild
+  fireEvent.change(input, {target: {value: 'c'}})
+  expect(spy).toHaveBeenCalledTimes(1)
 })
+
+/* eslint complexity:0 */
```
kentcdodds: Why were these changes necessary? It looks like a pure refactor.

src/index.js
```diff
@@ -59,14 +58,59 @@ function cleanupAtContainer(container) {
...
+  return simulate({
+    target: node,
+    type: eventName,
+    ...init,
+  })
+}
+
+// function written after some investigation here:
+// https://github.com/facebook/react/issues/10135#issuecomment-401496776
+function setNativeValue(element, value) {
```
kentcdodds: Maybe we should export this function from `dom-testing-library` so we can reuse that code here 🤔 

Would you like to make that PR there first? Should be a pretty simple one :)

src/index.js
```diff
@@ -59,14 +58,59 @@ function cleanupAtContainer(container) {
   mountedContainers.delete(container)
 }
 
+const originalChange = fireEvent.change
```
kentcdodds: Is it necessary to override the `change` event for this to work? I think we can remove the `change` event code here... I think that one works as-is

src/index.js
```diff
@@ -59,14 +58,59 @@ function cleanupAtContainer(container) {
...
+    setNativeValue(node, init.target.value)
+  }
+  return originalChange(node, init)
+}
+
+// probably there is a better way to get access to a component instance
+const [
+  getInstanceFromNode,
+] = ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Events
+const createReactEventSimulator = eventName => (node, init) => {
```
kentcdodds: Calling the prop directly sorta works, but I _think_ I'd rather come up with a way to fire the browser events that react is looking for when it triggers the callback.

src/__tests__/events.js
sivkoff: It's actually from your commit to original `pr/remove-simulate`, but agree we can drop these changes.

src/index.js
sivkoff: The same as above, but I can check if it's necessary.

src/index.js
sivkoff: Good idea! :)

src/index.js
sivkoff: OK, I'll dig into that deeper 😄

sivkoff:
Sure, I will do 👍

src/index.js
sivkoff: Added export in the `dom-testing-library`:
https://github.com/kentcdodds/dom-testing-library/pull/137

sivkoff:
OK, I got tests passing with using just native events 😄(thanks to @weyert's [comment](https://github.com/kentcdodds/react-testing-library/issues/206#issuecomment-432872702))
I'll commit it after some refactoring.

weyert:
Awesome!

kentcdodds:
I'm struggling to release this. There's something wrong with our automation. I may not get to it until after next week.

weyert:
The other ticket suggests it's a rights issue?

kentcdodds:
Sweet! Got this working!

PR #213, opened 2018-10-31
Link to external examples

skidding:
Not sure you'll find this useful. But I think the additional examples might help people getting started with RTL.


kentcdodds:
@antsmartian, good point. I'm pretty confident that not a ton will change in these packages and I'm not too concerned about keeping things being kept up to date. If they fall out of date, people can contribute to update them. Having a link to a resource is better than not :)

skidding:
Thanks!

PR #221, opened 2018-11-14
Update README.md

NoriSte:
I've aligned the documented signature of some functions in the README

I made that because I got confused at the beginning and I took some time to understand that they were just some examples (and not a real documentation).

- [x] Documentation
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -491,7 +491,9 @@ unmount()
 // your component has been unmounted and now: container.innerHTML === ''
 ```
 
-#### `getByLabelText(text: TextMatch, options: {selector = '*', exact = true, collapseWhitespace = true, trim = true}): HTMLElement`
+#### `getByLabelText(text: TextMatch, options): HTMLElement`
+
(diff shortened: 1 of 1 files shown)


NoriSte:
Done, I followed the same signing of the `dom-testing-library` readme. I avoided adding the types because the docs use the function as `h4` so too long title risks to make them unreadable.

Let me know what do you think or if you prefer I think better at changing the readme to make it "better"

README.md
```diff
@@ -491,7 +491,7 @@ unmount()
 // your component has been unmounted and now: container.innerHTML === ''
 ```
 
-#### `getByLabelText(text: TextMatch, options: {selector: string = '*'}): HTMLElement`
+#### `getByLabelText(text: TextMatch, options: {selector = '*', exact = true, collapseWhitespace = true, trim = true}): HTMLElement`
```
kentcdodds: This one is 👍

PR #246, opened 2018-12-15
Add ability to test component hydration

chrbala:
**What**: Add ability to test component hydration

**Why**: [Hydration](https://reactjs.org/docs/react-dom.html#hydrate) can be used for server side rendering. Hydration only attaches event listeners and does not change the props, even if a different tree is rendered. react-testing-library should allow testing of hydration for SSR.

**How**: A flag in render() that toggles hydration.

**Checklist**:

- [X] Documentation
- [X] Tests
- [X] Ready to be merged
- [X] Added myself to contributors table


kentcdodds:
Thanks! Could you add some documentation for this?

PR #259, opened 2019-01-04
fix(TS): add queries typings

chochihim:
**What**: Add typings for queries in RenderOptions

**Why**: It's missing. Also in a project I need to add some custom queries.

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      
The implementation is trickier than I thought. Any comment is appreciated. Also `Query` and `Queries` should probably be declared in `dom-testing-library`. I can create another PR once the implementation of this PR is accepted.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,31 +1,44 @@
-import {getQueriesForElement} from 'dom-testing-library'
+import {queries, BoundFunction} from 'dom-testing-library'
 
 export * from 'dom-testing-library'
 
-type GetsAndQueries = ReturnType<typeof getQueriesForElement>
+interface Query extends Function {
+  (container: HTMLElement, ...args): HTMLElement[] | HTMLElement | null
+}
 
-export interface RenderResult extends GetsAndQueries {
+interface Queries {
+  [T: string]: Query
+}
+
+export type RenderResult<Q extends Queries = typeof queries> = {
   container: HTMLElement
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -1,31 +1,44 @@
-import {getQueriesForElement} from 'dom-testing-library'
+import {queries, BoundFunction} from 'dom-testing-library'
 
 export * from 'dom-testing-library'
 
-type GetsAndQueries = ReturnType<typeof getQueriesForElement>
+interface Query extends Function {
+  (container: HTMLElement, ...args): HTMLElement[] | HTMLElement | null
```
joeporpeglia: @chochihim @kentcdodds heads up, the `...args` expression here causes TS builds to fail when `--noImplicitAny` is set.

```
Type error: Rest parameter 'args' implicitly has an 'any[]' type.  TS7019

    4 |
    5 | interface Query extends Function {
  > 6 |   (container: HTMLElement, ...args): HTMLElement[] | HTMLElement | null
      |                            ^
    7 | }
    8 |
    9 | interface Queries {
```

kentcdodds: 😬

Could you please open a PR to fix this @joeporpeglia.

joeporpeglia: Gotcha! #260

PR #262, opened 2019-01-11
BREAKING: bind queries to the container

brian-lives-outdoors:
**What**:

Bug fix: bind queries to the `container` in `render`

**Why**:

The docs state that ["the most important feature of `render` is that the queries from `dom-testing-library` are automatically returned with their first argument bound to the rendered container"](https://testing-library.com/docs/react-testing-library/api#queries).

But if no `container` was passed to `render`, the queries it was returning were bound to `document.body` (code [here](https://github.com/kentcdodds/react-testing-library/blob/14b567f87824b2a1b88d238793771e03d03937c4/src/index.js#L10-L15) and [here](https://github.com/kentcdodds/react-testing-library/blob/14b567f87824b2a1b88d238793771e03d03937c4/src/index.js#L50)).

**How**:

This PR ensures the queries are bound to the container, adds a corresponding unit test, and fixes [an older unit test](https://github.com/kentcdodds/react-testing-library/blob/14b567f87824b2a1b88d238793771e03d03937c4/src/__tests__/render.js#L45-L50) that was only passing because the queries were being incorrectly bound to `document.body`.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged
      
**Additional comments**

Bug was found while responding to [this StackOverflow issue](https://stackoverflow.com/q/54134434/10149510)


alexkrolick:
1. The current behavior is actually intentional to support querying Portals and other elements that render outside the root node. The docs could be rephrased to reflect this. https://github.com/kentcdodds/react-testing-library/issues/254#issuecomment-450034529
2. This would be a breaking change. If accepted we would need to add docs to explain how to use within() or unbound dom-testing queries to test modals, etc.

kentcdodds:
Hi @brian-lives-outdoors,

Thanks for the PR, but we will not be making this change. As @alexkrolick mentioned, this is intentional.

If you need this for some reason, then it's likely that you aren't cleaning up the document properly. Make sure to you're using `cleanup` (or `import 'react-testing-library/cleanup-after-each`).

Thanks again!

kentcdodds:
That said, we _would_ accept a PR to the docs to make them more clear here!

Issue #265, opened 2019-01-14
[types] RenderResult.container of type HTMLElement


SavePointSam:
The current typing of RenderResult.container leads to an error when trying to use container.firstChild.

Not all `HTMLElement`s have a `firstChild` property. Since we know [`container` should always be a `<div />`](https://github.com/kentcdodds/react-testing-library/blob/master/src/index.js#L14), it should be `HTMLDivElement`.

SavePointSam:
needed to add dom libs to my tsconfig.

```diff
{
  "compilerOptions": {
    "lib": [
+     "dom",
+     "dom.iterable",
      "esnext"
    ],
    ...
  },
}
```

PR #266, opened 2019-01-14
fix: correct container and baseElement typings

SavePointSam:
adjusts the typings of container and baseElement to better correlate to their real world application. fixes https://github.com/kentcdodds/react-testing-library/issues/265.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,3 +1,5 @@
+/// <reference lib="dom" />
+
 import {queries, BoundFunction} from 'dom-testing-library'
 
 export * from 'dom-testing-library'
@@ -11,8 +13,8 @@ interface Queries {
 }
 
 export type RenderResult<Q extends Queries = typeof queries> = {
-  container: HTMLElement
-  baseElement: HTMLElement
+  container: Node
+  baseElement: Node
   debug: (baseElement?: HTMLElement | DocumentFragment) => void
   rerender: (ui: React.ReactElement<any>) => void
   unmount: () => boolean


typings/index.d.ts
```diff
@@ -11,8 +11,8 @@ interface Queries {
 }
 
 export type RenderResult<Q extends Queries = typeof queries> = {
-  container: HTMLElement
-  baseElement: HTMLElement
+  container: HTMLDivElement | HTMLElement
+  baseElement: HTMLBodyElement | HTMLDivElement | HTMLElement
```
kentcdodds: I don't think this'll do what you want it to do. You'll still get the warning because `.firstChild` isn't on _all_ the possible types for the `container` or `baseElement`.

I think the real solution here is to remove `HTMLElement` in addition to your changes. Could you try that out by modifying this file in your `node_modules` first and say what happens?

SavePointSam: Yeah!

SavePointSam: Corrected!

kentcdodds:
Well, we're adding typings to `container` and `baseElement` both of which we create ourselves so we know their type and it's HTMLDivElement for both of them. I think _that_ would be the most accurate.

SavePointSam:
The input types are different than the outputs because [they are passed to the react renderer](https://github.com/kentcdodds/react-testing-library/blob/master/src/index.js#L25). When I dove into the renderer typings it just generically said `Element` which doesn't have a type for firstChild, that I could find.

SavePointSam:
Hmm, I tried it before and it wasn't working. I'll try with `HTMLDivElement` again later today.

julienw:
The hierarchy is ([according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement)):
`EventTarget -> Node -> Element -> HTMLElement -> HTMLDivElement`
`firstChild` is defined in `Node`.
I'm not sure what we're trying to fix here :) What's the specific error you get in #265?

SavePointSam:
@julienw

```
[ts] Property 'firstChild' does not exist on type 'HTMLElement'. [2339]
```

SavePointSam:
@kentcdodds, I'm still getting the does not exist error with `HTMLDivElement`. It only works if I use `Node`

```
[ts] Property 'firstChild' does not exist on type 'HTMLDivElement'. [2339]
```

SavePointSam:
Same with just `Element`

```
[ts] Property 'firstChild' does not exist on type 'Element'. [2339]
```

julienw:
Can you please show the full code that's failing for you?

SavePointSam:
I'm working on building a Codesandbox that reproduces the error. However, it appears to be working fine there. Double checking what's different about my development environment.

SavePointSam:
Ok, found my issue. Turns out my repo didn't list "dom" and "dom.iterable" in the lib section of my tsconfig. No type update is needed. I can look at updating the README to have a TypeScript settings section to avoid people hitting this in the future if you would like @kentcdodds.

kentcdodds:
Yeah, some docs updates would be appreciated!

PR #272, opened 2019-01-21
docs: add information about TypeScript

SavePointSam:
**What**:

Added explicit information about TypeScript in the README as well as a note about adding `dom` to the consumer `tsconfig.json`.

**Why**:

Since starting to use TypeScript, I've found it frustrating when packages I use do not have any information about TypeScript in their README as I have to go hunting. I also encountered the configuration issue outlined in the new documentation. It was hard to track down exactly why I was getting errors, so I thought it would be nice to have it right there.

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [x] Ready to be merged
      
### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,22 @@ You may also be interested in installing `jest-dom` so you can use
 > [**Docs**](https://testing-library.com/docs/react-testing-library/intro)
 
 
+## TypeScript
+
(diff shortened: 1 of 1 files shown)


kentcdodds:
Yeah, let's put this in the official docs plz. Thanks!

Issue #261, opened 2019-01-07
Add utility for testing custom hooks


kentcdodds:
### Describe the feature you'd like:

I think it would be great to add a utility to make testing custom hooks easier. Right now, [I recommend](https://www.youtube.com/watch?v=0e6WCQYg5tU&list=PLV5CVI1eNcJgCrPH_e6d57KRUTiDZgs0u) making a render prop component out of the hook and testing that component. I think we could do the same from within react-testing-library.

### Suggested implementation:

Here's a working implementation/example

```javascript
import React, {useState} from 'react'
import {render, cleanup} from 'react-testing-library'

afterEach(cleanup)

function useCounter({initialCount = 0, step = 1} = {}) {
  const [count, setCount] = useState(initialCount)
  const increment = () => setCount(c => c + step)
  const decrement = () => setCount(c => c - step)
  return {count, increment, decrement}
}

function testHook(useHook, props) {
  const RenderProp = ({children, ...rest}) => children(useHook(rest))
  const returnVal = {}
  render(
    <RenderProp {...props}>
      {val => {
        // may need some special treatment if the
        // return value is not an object of values...
        Object.assign(returnVal, val)
        return null
      }}
    </RenderProp>,
  )
  return returnVal
}


// how to use this:
test('renders counter', () => {
  const data = testHook(useCounter)
  data.increment()
  expect(data.count).toBe(1)
  data.decrement()
  expect(data.count).toBe(0)
})
```

### Describe alternatives you've considered:

- Not providing the utility at all.
- Calling it `renderHook` rather than `testHook`, or just about anything else. I'm not sure I like `testHook`...

### Teachability, Documentation, Adoption, Migration Strategy:

One really important consideration is that with `render` we recommend that people destructure just what they need rather than assigning the return value an actual variable name. With `testHook` however, if you destructure the return value it can have some pretty confusing issues (relating to the way JavaScript closures work, I explain it in the video I linked above). So all docs will need to call this out specially and all examples should not destructure anything from `testHook`.

We should also make sure to document the fact that this should typically only be used for hooks that either have a lot of edge cases that need special unit testing or highly reused hooks. Typically custom hooks should just be covered by testing the components that use them.

gnapse:
Sounds good, but I'm still skeptical that is can be advertised as a general solution to test custom hooks.

[...]

PR #274, opened 2019-01-22
Hooks Testing 🎣

donavon:
**What**:

Add a `testHook` utility

**Why**: 

You can't call a Hook directly as it must run in context of a React component.

**How**: 

`testHook` creates a component under-the-hood that calls back to the test where you can safely call the Hook. This so that the Hook can run within the context of a component, but set local deconstructed variables.

Here is an example of how to test a counter component.

```js
test('useCounter', () => {
  let count, increment
  testHook(() => ({count, increment} = useCounter({ initialCount: 2 })))

  expect(count).toBe(2)
  increment()
  expect(count).toBe(3)
})
```
(see above test running live in [CodeSandbox](https://codesandbox.io/s/6lqxvx42mz))

Any type of return types are supported in `hookTester`: objects, arrays, string, etc.

See discussion on #261 for more details.

**Checklist**:

- [x] Documentation - **under a separate PR into [docs repo](https://github.com/alexkrolick/testing-library-docs)**
- [x] Tests
- [x] Ready to be merged (**HOLD** until Hooks is released into prod with React 16.8)


gnapse:
Nice!

Some things off the top of my head that would be nice to add:

- More info about `testHook` in the README (I assume you knew this one and this is WIP)
- Type definitions for `testHook` in `typings/index.d.ts`
- Examples of tests using `testHook` under `examples/`
- Ideally some examples of using it would use something more than custom hooks based on just `useState` or `useReducer`. I'd love to see examples that showcase that testing a custom hook's side effects is also possible.

alexkrolick:
Implementation looks good. 

- For the Readme, just adding a link to a new `react-hooks` example should be sufficient
- Same example should go in the docs PR
- Once merged, @kentcdodds can add the example to the Codesandbox
- Also OK with iterating on the examples if we have at least one to start, even if it wraps a built-in hook. This can happen in the docs or Codesandbox.

donavon:
I've added `examples/use-counter.js` but in order to test it, it requires a build of React that supports Hooks. Thus... I had to change the `devDependencies` for `rest` and `react-dom` to `16.8.0-alpha.1`.

This will need to be bumped to the non-alpha in a few weeks when Hooks lands in 16.8.0.

So... this PR should be held and not merged until Hooks lands in production.

donavon:
README updated with link to `use-counter` example.

examples/__tests__/use-counter.js
```diff
@@ -0,0 +1,46 @@
+import {testHook, cleanup} from 'react-testing-library'
```
alexkrolick: This should be called `react-hooks` so the test is more about the APIs and less about the simulated component (similar to react-context etc.)

donavon: done.

package.json
```diff
@@ -48,15 +48,16 @@
     "jest-dom": "^2.0.4",
     "jest-in-case": "^1.0.2",
     "kcd-scripts": "^0.44.0",
-    "react": "^16.5.2",
-    "react-dom": "^16.5.2",
+    "react": "16.8.0-alpha.1",
+    "react-dom": "16.8.0-alpha.1",
```
alexkrolick: What if we put `^16.5.2 || 16.8.0-alpha.1` here?

src/index.js
```diff
@@ -51,6 +52,15 @@ function render(
   }
 }
 
+function TestHook({callback}) {
+  callback()
+  return null
+}
+
+const testHook = callback => {
```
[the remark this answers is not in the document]
gnapse: Glad to see I'm not alone on this one 😄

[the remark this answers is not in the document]
donavon: OK, let's do this...

Here's why I don't like using `function`. It's not immutable. 

If you declare a function like this, then it's able to be changed (intentionally or otherwise).

```js
> function test() { return 'test'; }
> test = 1;
```

If it were a `const`, you couldn't. I don't use `let` either (unless I can help it) for the same reason.

```js
> const test = () => 'test';
> test = 1;
VM619:1 Uncaught TypeError: Assignment to constant variable.
```

The other reason to use `const` with an arrow function assignment is that you can declare it as a *single statement* arrow function, and thus eliminate the `return` statement. Less visual clutter.

I'll change it as it's not my package, but I disagree.

alexkrolick:
> But the second 16.8.0 drops, all you have to do it re-run the job in Travis (it should pass, hopefully anyway 🤞) then merge the PR.

https://github.com/facebook/react/pull/14692

Better get this ready 🚦😉

kentcdodds:
That all sounds fine @donavon :+1:

donavon:
@alexkrolick 😱

examples/react-hooks.js
```diff
@@ -0,0 +1,10 @@
+import {useState} from 'react'
```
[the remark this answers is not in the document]
donavon: deferring per @kentcdodds 

see https://github.com/kentcdodds/react-testing-library/pull/274#issuecomment-457293235

typings/index.d.ts
```diff
@@ -40,6 +40,11 @@ export function render<Q extends Queries>(
   options: RenderOptions<Q>,
 ): RenderResult<Q>
 
+/**
+ * Renders a test component that calls back to the test.
+ */
+export function testHook(callback: () => void): void
```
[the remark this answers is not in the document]
donavon: the callback isn't called with anything and isn't required to return anything. i'm not a TS expert, so I'll defer, but this _should_ work.

src/index.js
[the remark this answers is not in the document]
donavon: Done. see 92b0c38

donavon:
dependencies updated and Travis build failed (as it should). with this, we should be ready to go! my work is done here. 😉

let me know if you need anything else.

examples/__tests__/react-hooks.js
```diff
@@ -0,0 +1,46 @@
+import {testHook, cleanup} from 'react-testing-library'
+
+import useCounter from '../react-hooks'
+
+afterEach(cleanup)
+
+test('accepts default initial values', () => {
+  let count
+  testHook(() => ({count} = useCounter()))
```
mAAdhaTTah: It might be nice if `testHook` could return the return value of `useCounter` rather than needing to fish it out ourselves like this. Thoughts?

alexkrolick: 🤔 

```jsx
function TestHook({callback}) {
  callback()
  return null
}

function testHook(callback) {
  let result
  let handleReturn = () => { result = callback() }
  render(<TestHook callback={handleReturn} />)
  return result
}
```

```js
// test.js
let {count, increment} = testHook(useCounter)
increment()
expect(count).toBe(1)
```

alexkrolick: @kentcdodds @donavon thoughts on that implementation?

donavon: what @kentcdodds said. this will only return the initial value returned by the hook. The code below won't work as `count` will still be `0`.

```js
let {count, increment} = testHook(() => useCounter(0))
increment()
expect(count).toBe(1)
```

So while it will work if your tests only test the initial value, it will add way too much confusion IMO. We should be this as-is.

mAAdhaTTah: Makes sense, thanks!

antsmartian:
@donavon Thank you so much for this PR. The code looks good to me, as Kent mentioned, we can fix some nits and get this asap when hooks are available! 👍

alexkrolick:
We should check out the new test utils API in https://github.com/reactjs/reactjs.org/pull/1629

kentcdodds:
The React team consulted with me about this. As soon as the new version lands I'll personally take on the work of updating things to work with hooks and their new APIs.

alexkrolick:
Ok great @kentcdodds 👍

kentcdodds:
Changed this to `next` branch so I can release this along with other changes to prepare for hooks.

kripod:
What happened to this feature? Unfortunately, I cannot find it anywhere...

bcarroll22:
:wave: you’ll wanna check out [react-hooks-testing-library](https://github.com/mpeyper/react-hooks-testing-library) if you’re testing custom hooks that you share between components. 

Otherwise, testing hooks via user interactions like clicks, etc. works just like anything else would with this library.

Good luck!

alexkrolick:
We should finish the PR to add that lib to the docs

tomstuart:
@kripod Here’s the information you’re missing: #302

PR #280, opened 2019-02-06
Release `testHook` and `act` support

kentcdodds:
**What**: Release `testHook` and `act` support

**Why**: Because React 16.8.0 was released!! Hooray!

**How**: See the individual PRs associated to this `next` branch for more info.

**Checklist**:

- [ ] Documentation (will happen on the docs site)
- [x] Tests
- [x] Ready to be merged


src/act-compat.js
```diff
@@ -0,0 +1,12 @@
+import {act as reactAct} from 'react-dom/test-utils'
+
+// act is supported react-dom@16.8.0
+// and is only needed for versions higher than that
+// so we do nothing for versions that don't support act.
+const act = reactAct || (cb => cb())
+
+function rtlAct(...args) {
+  return act(...args)
```
threepointone: you can actually polyfill it for older versions of react 
```jsx
function act(fn){
  ReactDOM.unstable_batchedUpdates(fn);
  ReactDOM.render(<div/>, document.createElement('div'));
}
```

you won't get the warnings, but something's better than nothing I suppose.

PR #282, opened 2019-02-06
Declare "module" in package.json

maisano:
**What**:

This change exposes the ESM version of the build, in the same fashion that the ES modules build is declared in `dom-testing-library`.

**Why**:

The ESM build is currently generated and distributed upon publish, though it is not exposed via `package.json`. My admittedly selfish use case for this is in having better heuristics around some linting rules (namely `import/named`).

**How**:
• Added the `module` field to `package.json`, pointing it at the path of the built artifact

**Checklist**:

- [ ] Documentation (N/A)
- [ ] Tests (N/A)
- [x] Ready to be merged

**&c.**:

Thanks for this library–it's great! Let me know if there's anything I overlooked or you would like to see changed.


kentcdodds:
Thanks!

PR #288, opened 2019-02-08
Fix act polyfill for React 15.x

### LINKED ISSUE #287: 5.5.3 breaks testing for React 15
- `react-testing-library` version: 5.5.3
- `react` version: 15.6.2
- `node` version: 10
- `npm` (or `yarn`) version: yarn 1.13

### Relevant code or config:

```js
import React from "react";
import { render } from "react-testing-library";
render(<div />);
```

### What you did:

Just recently installed the library and was trying to write my first tests. Everything was working fine until I updated to 5.5.3.

### What happened:

Trying to render anything in tests results in `ReactDOM.render(): Invalid component element.`

### Reproduction:

https://codesandbox.io/s/lxzrym3p9q

### Problem description:

This seems to be the culprit: https://github.com/kentcdodds/react-testing-library/blob/1d45c5c2ca1281187dc7735ac0e2f12137b157d3/src/act-compat.js#L10
Rendering `null` only works in React 16, not in 15.x.

### Suggested solution:

I'm actually not too sure why that line is necessary in the first place. Why is just calling `ReactDOM.unstable_batchedUpdates` not enough?

alexzherdev:
Fixes #287

**What**:

Restore compatibility with React 15.x.

**Why**:

The polyfill for `act` stopped working because `ReactDOM.render(null, div)` throws in React 15.x.

**How**:

Need to just render a dummy div in `act` polyfill.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged
      

alexzherdev:
Awesome! Out of curiosity, can you clarify what that render does?

kentcdodds:
It forces all effect callbacks to run synchronously.

alexzherdev:
I meant the `ReactDOM.render(<div />)`, not the `ReactDOM.unstable_batchedUpdates`. The test for the act polyfill passes without that `ReactDOM.render` call. No worries though.

Issue #289, opened 2019-02-09
Testing useEffect callback


Andrewmat:
### Describe the feature you'd like:

I've had this problem [on a personal project](https://github.com/Andrewmat/hooks/blob/master/src/hooks/useDocumentTitle.test.js) to try and create custom hooks.

The `useEffect` hook accepts a function as argument, and if this function returns another function, React will call it when the component unmounts

```javascript
useEffect(() => {
  subscribe(channel)
  return () => {
    // componentWillUnmount
    unsubscribe(channel)
  }
})
``` 

The testHook API already can test the useEffect call, but I couldn't find a way to test the callback on unmount.

### Suggested implementation:

The `testHook` function renders a wrapper component and already has access to the `unmount` function. My suggestion is to simply return it inside an object.
```jsx
function testHook(callback) {
  const { unmount } = render(<TestHook callback={callback} />)
  return { unmount }
}
```

This is simple to do, and I think it is clear too. Also it can be expand to other important return values when needed.

### Describe alternatives you've considered:

My first thought was to return the unmount function alone

```jsx
function testHook(callback) {
  const { unmount } = render(<TestHook callback={callback} />)
  return unmount
}
```

I did this to mimic the behavior of the `useEffect`. However, it didn't look good to me, as it would get all the usage of `testHook` return.  

### Teachability, Documentation, Adoption, Migration Strategy:

I've took the liberty to create a fork containing my sugestion.
https://github.com/Andrewmat/react-testing-library

Plus [tests](https://github.com/Andrewmat/react-testing-library/blob/master/src/__tests__/test-hook.js) and a [example](https://github.com/Andrewmat/react-testing-library/blob/master/examples/__tests__/react-hooks.js)

I've already tested in [my personal project](https://github.com/Andrewmat/hooks/blob/testing-unmount/src/hooks/useDocumentTitle.test.js) using `yarn link` and it worked.

![captura de tela de 2019-02-08 21-38-14](https://user-images.githubusercontent.com/5133846/52512801-08b80800-2bee-11e9-97b9-e097913188ca.png)

If this feature is considered, I'd be glad to make a pull request of it.

kentcdodds:
Why don't we just return everything?

Andrewmat:
I think it would be an overkill, and since they all control the React element, I assume it *could* encourage some people to make tests basing on the TestHook component instead of the hook.
Although, I see the value of `rerender` to check the deps array of `useEffect`

[...]

PR #290, opened 2019-02-09
Modifies testHook to return unmount function

Andrewmat:
This PR solves #289, in which I asked to have a way to test the `useEffect` callback.

**What**:
Adds a way to test hooks behavior when component unmounts.

**Why**:
To test `useEffect` callbacks, which are called when component unmounts.

**How**:
Modifies `testHook` to return an object with an exposed `unmount` function in it.
As said in the issue #289, I didn't expose anything else to avoid encouraging unfocused tests. In the future, if someone asks for them, they could be included.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

I also modified the file `examples/react-hooks` so `useCounter` is not a default export, and added `useDocumentTitle` as example.
I am not confident about the `useDocumentTitle` though, as its implementation is quirky and not a good example (*ironically*). But I couldn't think of another simple use of `useEffect` and its callback.


examples/__tests__/react-hooks.js
```diff
@@ -6,53 +6,76 @@
  */
```
kentcdodds: I'm guessing you had to skip the pre-commit hook or something. This should have auto-formatted when you commit it. Could you run `npx kcd-scripts format` please?

src/index.js
```diff
@@ -67,7 +67,8 @@ function TestHook({callback}) {
 }
 
 function testHook(callback) {
-  render(<TestHook callback={callback} />)
+  const { unmount } = render(<TestHook callback={callback} />)
```
kentcdodds: Let's also return `rerender` as well I think.

examples/__tests__/react-hooks.js
Andrewmat: I ran it and it added a linebreak in `ISSUE_TEMPLATE.md` and `mock.react-transition-group.js`. Should I also add it to the commit?
captura de tela de 2019-02-10 12-48-41
captura de tela de 2019-02-10 12-48-23

FredyC:
Would be nice to update typescript definition when changing API. 🙊

Andrewmat:
I haven't mess with TS typing for a long time now. Can you check if this is correct?
https://github.com/Andrewmat/react-testing-library/blob/master/typings/index.d.ts

kentcdodds:
Looks correct to me.

Andrewmat:
Should I also add some guidance inside CONTRIBUTING.md?
```
If your PR introduced some changes in the API, you are more than welcome to modify the Typescript type definition to reflect those changes. Just modify the `/typings/index.d.ts` file accordingly. If you have never seen Typescript definitions before, you can read more about it in their [documentation pages](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html)
```
I thought about adding a new section in "Commiting and Pushing changes"

kentcdodds:
Sure :+1:

thealjey:
wow, mind absolutely blown

I love this library

PR #340, opened 2019-04-01
fix(act): add try/catch around react-dom/test-utils require

### LINKED ISSUE #315: Cannot find module 'react-dom/test-utils' from 'act-compat.js'
- `react-testing-library` version: 6.0.0
- `react` version: 15.4.1
- `node` version: 8.11.3
- `npm` (or `yarn`) version: 1.12.3

### Relevant code or config:

```js
import 'react-testing-library/cleanup-after-each';
```

### What you did:

Just imported the `react-testing-library/cleanup-after-each`

### What happened:

Got this error `Cannot find module 'react-dom/test-utils' from 'act-compat.js'`

### Problem description:

The `react-addons-test-utils` was moved to `react-dom/test-utils` in the version `15.5`, that path doesn't work in `react-dom <15.5`

### Suggested solution:

Import from `react-addons-test-utils` or `react-dom/lib/ReactTestUtils` for `react-dom 15.4.x` versions, see this line
https://github.com/kentcdodds/react-testing-library/blob/master/src/act-compat.js#L3

kentcdodds:
**What**: add try/catch around react-dom/test-utils require

**Why**: Closes #315

**How**: replace import with require so we can make it happen dynamically

**Checklist**:

- [x] Documentation added to the [docs site](https://github.com/alexkrolick/testing-library-docs) N/A
- [x] Tests N/A
- [x] Ready to be merged
      
- [x] Added myself to contributors table N/A

### DIFF
--- a/src/act-compat.js
+++ b/src/act-compat.js
@@ -1,6 +1,12 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
-import {act as reactAct} from 'react-dom/test-utils'
+
+let reactAct
+try {
+  reactAct = require('react-dom/test-utils').act
+} catch (error) {
+  // ignore, this is to support old versions of react
+}
 
 // act is supported react-dom@16.8.0
 // so for versions that don't have act from test utils


kentcdodds:
If I could get one review that would be stellar 👍

src/act-compat.js
```diff
@@ -1,6 +1,12 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
-import {act as reactAct} from 'react-dom/test-utils'
+
+let reactAct
+try {
+  reactAct = require('react-dom/test-utils').act
```
alexkrolick: Will this throw if act is not defined but test-utils is?

kentcdodds: No it wont, but we'll handle it fine anyway thanks to the `||` below 👍

PR #343, opened 2019-04-05
feat(act): Support async act 🎉

### LINKED ISSUE #281: [5.5.0] Getting warning about "act" anyway
Thanks for the fast work Kent, but seems something is still off. I've updated to 5.5.0 and tests are still generating same act warnings. Am I missing something here? Do we need to actually use `act` inside the tests?

https://travis-ci.org/mobxjs/mobx-react-lite/builds/489613981

It's curious that I've actually removed a bunch of `flushEffects` calls which were necessary before and tests are passing just fine. Clearly `act` helped with that. And yet those warnings shouldn't be there. 🤔

kentcdodds:
**What**: Add support for async `act`

**Why**: Closes #281

**How**: Unfortunately, we need to do a bit of attempts at polyfilling to ensure we support react-dom@16.8.0 for the foreseeable future. This means that we not only have to polyfill `act` itself, but also polyfill the async support of `act`. So I've done my best to detect whether `act` supports async. I feel pretty good about this solution. I'd rather this than detecting the version number anyway.

**Checklist**:

- [x] Documentation added to the [docs site](https://github.com/alexkrolick/testing-library-docs) N/A
- [x] Tests N/A
- [x] Ready to be merged
      
- [x] Update the dom-testing-library version once this is released: https://github.com/kentcdodds/dom-testing-library/pull/236

### DIFF
--- a/src/__tests__/old-act.js
+++ b/src/__tests__/old-act.js
@@ -1,0 +1,45 @@
+import {asyncAct} from '../act-compat'
+
+jest.mock('react-dom/test-utils', () => ({
+  act: cb => {
+    const promise = cb()
+    return {
+      then() {
+        console.error('blah, do not do this')
+        return promise
+      },
+    }
+  },
+}))
+
+test('async act works even when the act is an old one', async () => {
+  jest.spyOn(console, 'error').mockImplementation(() => {})
+  const callback = jest.fn()
+  await asyncAct(async () => {
+    await Promise.resolve()
+    await callback()
+  })
+  expect(console.error.mock.calls).toMatchInlineSnapshot(`
+Array [
+  Array [
+    "It looks like you're using a version of react-dom that supports the \\"act\\" function, but not an awaitable version of \\"act\\" which you will need. Please upgrade to at least react-dom@16.9.0 to remove this warning.",
+  ],
+]
+`)
+  expect(callback).toHaveBeenCalledTimes(1)
+
+  // and it doesn't warn you twice
+  callback.mockClear()
+  console.error.mockClear()
+
+  await asyncAct(async () => {
+    await Promise.resolve()
+    await callback()
+  })
+  expect(console.error).toHaveBeenCalledTimes(0)
+  expect(callback).toHaveBeenCalledTimes(1)
+
+  console.error.mockRestore()
+})
+
+/* eslint no-console:0 */
--- a/src/act-compat.js
+++ b/src/act-compat.js
@@ -2,8 +2,23 @@ import React from 'react'
 import ReactDOM from 'react-dom'
 
 let reactAct
+let actSupported = false
+let asyncActSupported = false
 try {
   reactAct = require('react-dom/test-utils').act
+  actSupported = reactAct !== undefined
+
+  const originalError = console.error
+  let errorCalled = false
+  console.error = () => {
+    errorCalled = true
+  }
+  console.error.calls = []
+  reactAct(() => ({then: () => {}})).then(/* istanbul ignore next */ () => {})
+  if (!errorCalled) {
+    asyncActSupported = true
+  }
+  console.error = originalError
 } catch (error) {
   // ignore, this is to support old versions of react
 }
@@ -19,8 +34,28 @@ function actPolyfill(cb) {
 
 const act = reactAct || actPolyfill
 
-function rtlAct(...args) {
-  return act(...args)
+let youHaveBeenWarned = false
+// this will not avoid warnings that react-dom 16.8.0 logs for triggering
+// state updates asynchronously, but at least we can tell people they need
+// to upgrade to avoid the warnings.
+async function asyncActPolyfill(cb) {
+  if (!youHaveBeenWarned && actSupported) {
+    // if act is supported and async act isn't and they're trying to use async
+    // act, then they need to upgrade from 16.8 to 16.9.
+    // This is a seemless upgrade, so we'll add a warning
+    console.error(
+      `It looks like you're using a version of react-dom that supports the "act" function, but not an awaitable version of "act" which you will need. Please upgrade to at least react-dom@16.9.0 to remove this warning.`,
+    )
+    youHaveBeenWarned = true
+  }
+  await cb()
+  // make all effects resolve after
+  act(() => {})
 }
 
-export default rtlAct
+const asyncAct = asyncActSupported ? reactAct : asyncActPolyfill
+
+export default act
+export {asyncAct}
+
+/* eslint no-console:0 */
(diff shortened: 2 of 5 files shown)


kentcdodds:
Note: That even having an outdated dom-testing-library works fine. I didn't think it would, but it makes sense now so that's cool :)

src/act-compat.js
```diff
@@ -19,8 +34,28 @@ function actPolyfill(cb) {
...
+  if (actSupported) {
+    // if act is supported and async act isn't and they're trying to use async
+    // act, then they need to upgrade from 16.8 to 16.9.
+    // This is a seemless upgrade, so we'll add a warning
+    console.error(
+      `It looks like you're using a version of react-dom that supports the "act" function, but not an awaitable version of "act" which you will need. Please upgrade to at least react-dom@16.9.0 to remove this warning.`,
+    )
+  }
+  // make all effects resolve before
+  act(() => {})
```
threepointone: we went back and forth on this, and decided not to flush in the beginning. is there a reason you added this here?

src/act-compat.js
```diff
@@ -19,8 +34,28 @@ function actPolyfill(cb) {
...
-  return act(...args)
+// this will not avoid warnings that react-dom 16.8.0 logs for triggering
+// state updates asynchronously, but at least we can tell people they need
+// to upgrade to avoid the warnings.
+async function asyncActPolyfill(cb) {
+  if (actSupported) {
+    // if act is supported and async act isn't and they're trying to use async
+    // act, then they need to upgrade from 16.8 to 16.9.
+    // This is a seemless upgrade, so we'll add a warning
+    console.error(
```
threepointone: maybe you'd want to use a flag so this warning happens just once?

src/act-compat.js
kentcdodds: Thanks for that! I wasn't sure whether this was correct 😅 I'll remove that.

src/act-compat.js
kentcdodds: Oh! Duh. I meant to do that.

src/__tests__/old-act.js
```diff
@@ -0,0 +1,45 @@
...
+  expect(console.error.mock.calls).toMatchInlineSnapshot(`
+Array [
+  Array [
+    "It looks like you're using a version of react-dom that supports the \\"act\\" function, but not an awaitable version of \\"act\\" which you will need. Please upgrade to at least react-dom@16.9.0 to remove this warning.",
+  ],
+]
+`)
+  expect(callback).toHaveBeenCalledTimes(1)
+
+  // and it doesn't warn you twice
```
kentcdodds: 😘 @threepointone

kentcdodds:
This looks good to go from my perspective. I'll leave it open for a while for folks to review if they want.

src/act-compat.js
```diff
@@ -19,8 +34,28 @@ function actPolyfill(cb) {
 
 const act = reactAct || actPolyfill
 
-function rtlAct(...args) {
-  return act(...args)
+let youHaveBeenWarned = false
```
alexkrolick: 😂

kentcdodds:
Saweeet! Off it goes.

alexkrolick:
Do we have to `await` fireEvent now since it's wrapped? Or does it work synchronously too?

@kentcdodds

threepointone:
you'll have to wrap `fireEvent` with `act()` if you have anything async going on. 
```jsx
await act(async () => {
  fireEvent(...)
})
```

you could write a helper of course
```jsx
async function fireEventAsync(){
  await act(async () => {
    fireEvent(...)
  })
}
// ...
await fireEventAsync()
```

alexkrolick:
@threepointone it is wrapped https://github.com/kentcdodds/react-testing-library/blob/master/src/index.js#L100

Wondering if this changes the required usage to need await

kentcdodds:
We don't need a `fireEventAsync`, people can use the `findBy*` queries or one of the `wait*` utilities which are now all wrapped.

`fireEvent` was already wrapped. No need to `await` an `act` that is not async.

Bottom line: We're solid.

threepointone:
Oh haha I forgot *facepalm* your lib is awesome

FredyC:
@kentcdodds Can you please elaborate on how is `findBy` useful for firing events? I am confused :)

kentcdodds:
`findBy*` is _not_ useful for firing events, but you could do it for finding an element that loads asynchronously _after_ an event is fired:

```
fireEvent.click(loadGreeting)
const greetingNode = await findByText(/hello/i)
```

FredyC:
Oh wow, this is going to take some time to process. So far I've been using `getBy*` most of the time and did not even know there is some `findBy*`.

So to be clear, if I am doing something else besides finding elements, eg. `expect(container.textContext)...`  AND there is some async code then I *have to* wrap `fireEvent` into extra `await act`? If that's the case, it's going to be fairly confusing, to be honest. Especially when someone else looks at those tests and some will be wrapped and some not.

kentcdodds:
No. You should _never_ have to wrap a `fireEvent` call in `act`. We do that for you. You shouldn't have to wrap _any_ code in `act` 99% of the time, so long as you're using react-testing-library utilities, you should be good.

kentcdodds:
Also, `findBy*` is only a few weeks old, so it's understandable that you don't know about it. Check the docs :)

FredyC:
Well, so I made [some convoluted example](https://codesandbox.io/s/6jpr2ow8mk?fontsize=14&moduleview=1&previewwindow=tests), but I am doing similar stuff fairly often. Sadly it's not passing and do wonder what's is the proper way out of it.

```jsx
const TestComponent = () => {
  const [state, setState] = React.useState(1)
  const onClick = () => {
    setImmediate(() => {
      setState(i => i + 1)
    })
  }

  return (
    <>
      <button onClick={onClick}>GO</button>
      <div>{state}</div>
    </>
  )
}

test('nay', () => {
  const { getByText, container } = render(<TestComponent />)
  fireEvent.click(getByText('GO'))
  expect(container.textContent).toBe('GO2')
})
```

There is also an attempt to solve it with `findBy*` which seems to work most of the time, but for some reason is flaky as it has also failed a couple of times when re-running it 🤔 

_Sorry, I should have probably made a new issue, but I am kinda tired here and heading to bed. That might be also a reason why I missed something there 🤷‍♂️_

kentcdodds:
That's pretty weird. `findBy*` queries should reject if no element is found. Could you dig a little deeper and open an issue with your findings?

FredyC:
Sorry, have no capacity for diving deep in there right now. I don't think I will ever need `findBy*` as it feels weird to me. So far I had no need to wait for some element to appear.

I've updated the CodeSandbox with two passing tests and one failing (`findByTitle`). Dig into it or leave it, up to you ;)

https://codesandbox.io/s/6jpr2ow8mk

PR #394, opened 2019-06-24
fix: Specify baseElement without container

eps1lon:
**What**:

Fix `baseElement` being overridden if no `container` was provided

**Why**:

I'm having a hard time grasping the difference between `baseElement` and `container` in the current implementation:
> The containing DOM node where your React Element is rendered in the container.

-- https://testing-library.com/docs/react-testing-library/api#baseelement-1

Considering the default values I imagined it as 
```html
<base-element>
  <container>
    <Component />
  </container>
</base-element>
```

The `base-element` acts as a long living container that will hold every rendered container but can be isolated from the rest of the document.

What I'm basically trying to achieve is that the queries don't look inside document.body but rather only in the DOM tree rendered by the component. It's not so much about portals but other react trees living in the same document that should be ignored.

**How**:

**Checklist**:

      [docs site](https://github.com/alexkrolick/testing-library-docs) Note: Should I rather update the documentation?
- [x] Tests
- ~[ ] Typescript definitions updated~
- [ ] Ready to be merged Note: If this is indeed a bug, yes. Otherwise I repurpose this as a feature request

### DIFF
--- a/src/__tests__/render.js
+++ b/src/__tests__/render.js
@@ -113,3 +113,44 @@ test('renders options.wrapper around node', () => {
 </div>
 `)
 })
+
+describe('baseElement', () => {
+  let baseElement
+  beforeAll(() => {
+    baseElement = document.createElement('div')
+    document.body.appendChild(baseElement)
+  })
+
+  afterAll(() => {
+    baseElement.parentNode.removeChild(baseElement)
+  })
+
+  it('can take a custom element for isolation', () => {
+    function DescribedButton({title: description, ...buttonProps}) {
+      return (
+        <React.Fragment>
+          <button {...buttonProps} aria-describedby="tooltip" />
+          {ReactDOM.createPortal(
+            <div id="tooltip" role="tooltip">
+              {description}
+            </div>,
+            document.body,
+          )}
+        </React.Fragment>
+      )
+    }
+
+    const {getByRole, getByText} = render(
+      <DescribedButton description="this descripton is hidden from rtl">
+        Click me
+      </DescribedButton>,
+      {baseElement},
+    )
+
+    expect(getByText('Click me')).toBeTruthy()
+    expect(document.querySelector('[role="tooltip"]')).toBeTruthy()
+    expect(() => getByRole('tooltip')).toThrow(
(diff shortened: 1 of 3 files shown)


src/__tests__/render.js
```diff
@@ -113,3 +113,44 @@ test('renders options.wrapper around node', () => {
 </div>
 `)
 })
+
+describe('baseElement', () => {
```
kentcdodds: I really prefer to not use any nesting. This is all for just one test, please instead move the contents of beforeAll and afterAll to the inside of the test and use the `test` global rather than the `describe` and `it` globals.

src/__tests__/render.js
```diff
@@ -113,3 +113,44 @@ test('renders options.wrapper around node', () => {
...
+    baseElement = document.createElement('div')
+    document.body.appendChild(baseElement)
+  })
+
+  afterAll(() => {
+    baseElement.parentNode.removeChild(baseElement)
+  })
+
+  it('can take a custom element for isolation', () => {
+    function DescribedButton({title: description, ...buttonProps}) {
```
kentcdodds: Is it necessary that this component be so complex? I prefer to make the component simpler/contrived to get the point across. I believe that this test would pass with or without the `baseElement`... Can we have a test that actually requires the `baseElement`?

eps1lon: I'll give it a try to make this more meaningful. It does, however, fail without the fix applied or without the baseElement passed.

src/__tests__/render.js
eps1lon: This breaks two other test that assume an empty body. The nesting is for isolation. What is the problem with that? The alternative is to create a new test which is worse IMO.

kentcdodds: I don't like the direction that nesting puts us in. It's a slippery slope IMO. Maybe I'm just too scarred by past testbases. I'm much more willing to create a new test file than nesting tests.

Another thing we could do is just take the `beforeAll` and `afterAll` and put them inside the test. That would avoid the problems altogether. Could you do that please?

eps1lon:
Hope that makes more sense.

PR #396, opened 2019-06-25
fix(TS): export `Queries` interface

### LINKED ISSUE #395: Export `Queries` interface
- `react-testing-library` version: 8.0.2
- `react` version: 16.8.6
- `node` version: 8.9.0
- `yarn` version: 1.16.0

### Relevant code or config:

```ts
export function render<Q extends Queries>(
  ui: React.ReactElement<any>,
  options: MyCustomRenderOptions<Q>,
): RenderResult<Q>
```

### What you did:

I would like to provide typings for a custom `render` function https://testing-library.com/docs/react-testing-library/setup#custom-render where I allow an extra configuration option (`theme` in my particular case). Now I'm not a TS guru, but my assumption is that I would have to copy the current definition of `render` and customise `options`.

### What happened:

> Module '"bla/node_modules/@testing-library/react/typings/index"' has no exported member 'Queries'. Did you mean 'queries'?

### Problem description:

See above.

### Suggested solution:
I think `Queries` should be exported. https://github.com/microsoft/web-build-tools/blob/master/apps/api-extractor/README.md says

> **Missing exports:** Suppose the **awesome-widgets** package exports an API function `AwesomeButton.draw()` that requires a parameter of type `DrawStyle`, but you forgot to export this enum.  Things seem fine at first, but when a developer tries to call that function, they discover that there's no way to specify the `DrawStyle`.

I'm happy to raise a PR for this if this request is deemed reasonable.

https://github.com/testing-library/react-testing-library/blob/8e99f3cf99b76afbf2b20993745b802def3bb6c9/typings/index.d.ts#L15

```diff
-interface Queries {
+export interface Queries {
```

NMinhNguyen:
**What**: Export `Queries` interface to make it easier to define a custom `render`, e.g.

```js
import { render } from '@my-design-system/test-utils'
// Pass in a custom theme which the custom `render` wires up to a `ThemeProvider`
render(<MyDesignSystemButton />, { theme: { someColor: 'blue' } })
```

**Why**: Closes #395.

**How**: Export `Queries` interface which is referenced by `render` in `index.d.ts`.

**Checklist**:

- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -12,7 +12,7 @@ interface Query extends Function {
     | null
 }
 
-interface Queries {
+export interface Queries {
   [T: string]: Query
 }
 


kentcdodds:
@all-contributors please add @NMinhNguyen for code

NMinhNguyen:
> That seems reasonable to me! Thanks!

Actually @kentcdodds upon further investigation, it seems like it was some kinda misconfiguration issue at work (can't seem to repro, however) where it threw an error using a much older version of RTL (5.5.3), and it doesn't seem to throw on 8.0.2.. When comparing the two `index.d.ts` files, I saw that `Queries` wasn't being exported in either version, so just assumed the issue was going to be present in 8.0.2 too - fixing 5.5.3 in `node_modules` made it go away. Oh well, I guess it doesn't hurt to have this keyword 🤷‍♂

PR #407, opened 2019-07-22
fix(act): async act detection

threepointone:
The previous version of async act detection left an open hanging act scope, which broke tests and expectations. This PR delays the detection until it's been called at least once.

I tested this with 16.8.6, and 0.0.0-b4178af81.

### DIFF
--- a/src/__tests__/old-act.js
+++ b/src/__tests__/old-act.js
@@ -1,4 +1,14 @@
-import {asyncAct} from '../act-compat'
+let asyncAct
+
+beforeEach(() => {
+  jest.resetModules()
+  asyncAct = require('../act-compat').asyncAct
+  jest.spyOn(console, 'error').mockImplementation(() => {})
+})
+
+afterEach(() => {
+  console.error.mockRestore()
+})
 
 jest.mock('../react-dom-16.9.0-is-released', () => ({
   reactDomSixteenPointNineIsReleased: true,
@@ -6,30 +16,40 @@ jest.mock('../react-dom-16.9.0-is-released', () => ({
 
 jest.mock('react-dom/test-utils', () => ({
   act: cb => {
-    const promise = cb()
+    cb()
     return {
       then() {
-        console.error('blah, do not do this')
-        return promise
+        console.error(
+          'Warning: Do not await the result of calling ReactTestUtils.act(...), it is not a Promise.',
+        )
       },
     }
   },
 }))
 
 test('async act works even when the act is an old one', async () => {
-  jest.spyOn(console, 'error').mockImplementation(() => {})
   const callback = jest.fn()
   await asyncAct(async () => {
+    console.error('sigil')
     await Promise.resolve()
     await callback()
+    console.error('sigil')
   })
   expect(console.error.mock.calls).toMatchInlineSnapshot(`
-Array [
-  Array [
-    "It looks like you're using a version of react-dom that supports the \\"act\\" function, but not an awaitable version of \\"act\\" which you will need. Please upgrade to at least react-dom@16.9.0 to remove this warning.",
-  ],
-]
-`)
+        Array [
+          Array [
+            Array [
+              "sigil",
+            ],
+          ],
+          Array [
+            "It looks like you're using a version of react-dom that supports the \\"act\\" function, but not an awaitable version of \\"act\\" which you will need. Please upgrade to at least react-dom@16.9.0 to remove this warning.",
+          ],
+          Array [
+            "sigil",
+          ],
+        ]
+    `)
   expect(callback).toHaveBeenCalledTimes(1)
 
   // and it doesn't warn you twice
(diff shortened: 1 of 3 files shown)


src/act-compat.js
```diff
@@ -2,29 +2,8 @@ import React from 'react'
 import ReactDOM from 'react-dom'
 import {reactDomSixteenPointNineIsReleased} from './react-dom-16.9.0-is-released'
 
-let reactAct
-let actSupported = false
-let asyncActSupported = false
-try {
```
threepointone: I'm not sure why this try catch block was here. was it to catch the error if reactAct was undefined? I'm guessing so.

src/__tests__/old-act.js
```diff
@@ -6,11 +6,12 @@ jest.mock('../react-dom-16.9.0-is-released', () => ({
 
 jest.mock('react-dom/test-utils', () => ({
   act: cb => {
-    const promise = cb()
```
threepointone: 16.8.6's act actually does not return the promise.

src/__tests__/old-act.js
```diff
@@ -6,11 +6,12 @@ jest.mock('../react-dom-16.9.0-is-released', () => ({
...
+    cb()
     return {
       then() {
-        console.error('blah, do not do this')
-        return promise
+        console.error(
+          'Warning: Do not await the result of calling ReactTestUtils.act(...), it is not a Promise.',
```
threepointone: matching the error is now essential to this test.

src/act-compat.js
```diff
@@ -38,32 +17,96 @@ function actPolyfill(cb) {
...
+              act(() => {})
+              resolve()
+            })
+          })
+        }
+      })
+    } else if (isAsyncActSupported === false) {
+      // use the polyfill directly
+      let result
+      act(() => {
```
threepointone: maybe this should be batchedUpdates

.watchmanconfig
```diff
@@ -0,0 +1 @@
+{}
```
kentcdodds: Any chance we can avoid this file? I've never needed it before.

src/__tests__/old-act.js
```diff
@@ -4,32 +4,92 @@ jest.mock('../react-dom-16.9.0-is-released', () => ({
...
-      then() {
-        console.error('blah, do not do this')
-        return promise
-      },
-    }
-  },
-}))
-
 test('async act works even when the act is an old one', async () => {
+  jest.mock('react-dom/test-utils', () => ({
```
kentcdodds: I really prefer to not put `jest.mock` inside a test. It makes people think that the mock only applies for that test and not the others which is not correct. Could we leave this at the top level as before?

src/__tests__/old-act.js
```diff
@@ -4,32 +4,92 @@ jest.mock('../react-dom-16.9.0-is-released', () => ({
...
+  expect(console.error).toHaveBeenCalledTimes(0)
+  expect(callback).toHaveBeenCalledTimes(1)
+
+  console.error.mockRestore()
+
+  jest.unmock('react-dom/test-utils')
+})
+
+test('should not warn with a version that resolves', async () => {
+  jest.mock('react-dom/test-utils', () => ({
```
kentcdodds: I'm pretty sure that putting `jest.mock` and `jest.unmock` throughout this file would require that you do `jest.resetModules()` and `require` the modules in each test... 🤔

I do it that way sometimes, but the way I normally solve this problem is by putting the tests that require different mocks in separate files or by making the thing I want mocked a mock function and using `mockImplementationOnce` for it.

In this case, one file for old react (that doesn't have act) and another for one that does which behaves differently (via `mockImplementationOnce`) seems like the right solution.

.watchmanconfig
threepointone: yeah I guess. a newer version of watchman will error without this file, but that's only internal now. will remove.

src/__tests__/old-act.js
threepointone: yeah I might not be doing the right thing with these tests. mockImplementationOnce is a good lead.

.watchmanconfig
kentcdodds: Can you talk to those wonderful internal people and say that I will be quite sad if this is the case in the future 😅

threepointone: yeah I think they're aware. I'll check.

src/act-compat.js
```diff
@@ -38,32 +17,101 @@ function actPolyfill(cb) {
...
+          cbReturn = cb()
+          return cbReturn
+        })
+
+        result.then(() => {
+          console.error = originalConsoleError
+          // if it got here, it means async act is supported
+          isAsyncActSupported = true
+          resolve()
+        }, reject)
```
gaearon: If it fails do you restore console.error?

threepointone: good catch, it doesn't! fixing.

threepointone: fixed, expanded tests to include it too

src/act-compat.js
threepointone: it shouldn't actually.

kentcdodds:
I'd love to have one other approval on this before I merge 👍

kentcdodds:
P.S. Thanks a ton Sunil!

src/act-compat.js
```diff
@@ -38,32 +17,114 @@ function actPolyfill(cb) {
...
+          if (
+            args[0].indexOf(
+              'Warning: Do not await the result of calling ReactTestUtils.act',
+            ) === 0
+          ) {
+            // v16.8.6
+            isAsyncActSupported = false
+          } else if (
+            args[0].indexOf(
+              'Warning: The callback passed to ReactTestUtils.act(...) function must not return anything',
```
alexkrolick: I guess we are assuming these strings are stable? There's no error codes or something we could use?

alexkrolick: I guess these are already released, and no more 16.8.x releases are planned, so 🤷‍♂

src/__tests__/no-act.js
```diff
@@ -7,3 +13,69 @@ test('act works even when there is no act from test utils', () => {
...
+  console.error.mockClear()
+
+  await asyncAct(async () => {
+    await Promise.resolve()
+    await callback()
+  })
+  expect(console.error).toHaveBeenCalledTimes(0)
+  expect(callback).toHaveBeenCalledTimes(1)
+
+  console.error.mockRestore()
```
alexkrolick: minor, but, technically shouldn't this be in a before or after hook so it doesn't get left in a weird state if part of the test fails?

kentcdodds: True. Let's do this:

```javascript
beforeEach(() => {
  jest.spyOn(console, 'error').mockImplementation(() => {})
})

afterEach(() => {
  console.error.mockRestore()
})
```

threepointone: done

kentcdodds:
@all-contributors please add @threepointone for code and tests

kentcdodds:
@all-contributors, please add @gaearon for reviews

bcarroll22:
Little late to the game, sorry for that, I'm trying to figure out if this is something worth porting to NTL?

kentcdodds:
Probably

ghengeveld:
@threepointone This change is causing an issue: https://github.com/testing-library/react-testing-library/issues/468

PR #427, opened 2019-08-06
await act(...) inside cleanup-after-each

threepointone:
**What**:

This PR adds an async `act()` call to `cleanup-after-each`. 

**Why**:

Tests, especially those with async logic, could have hanging microtasks or React work that could leak into the next test. 

**How**:

This PR
- adds `cleanup-after-each.js` (so I can write tests for it)
- awaits an `act(async () => {})` call before unmounting containers
- And a test for the same

**Some possible Q&A**:

- Why not do a sync `act()` in `cleanup()`?: If peeps are using react-testing-library, it's suuuper unlikely they'll have hanging sync effects/updates. Decided not to add code without a good reason.

(bonus: fixes a lint violation in act-compat.js)

**Checklist**:

- [x] Documentation added to the [docs site](https://github.com/alexkrolick/testing-library-docs) - **NA**
- [x] Tests
- [x] Typescript definitions updated - **NA**
- [x] Ready to be merged

### DIFF
--- a/cleanup-after-each.js
+++ b/cleanup-after-each.js
@@ -1,1 +1,3 @@
-afterEach(require('./dist').cleanup)
+afterEach(() => {
+  return require('./dist/cleanup-async')()
+})
--- a/src/__tests__/cleanup-after-each.js
+++ b/src/__tests__/cleanup-after-each.js
@@ -1,0 +1,32 @@
+import React from 'react'
+import {render} from '../index'
+import cleanupAsync from '../cleanup-async'
+
+afterEach(() => {
+  return cleanupAsync()
+})
+
+const log = []
+let ctr = 0
+
+function App() {
+  async function somethingAsync() {
+    await null
+    log.push(ctr++)
+  }
+  React.useEffect(() => {
+    somethingAsync()
+  }, [])
+  return 123
+}
+
+it('cleanup-after-each does not leave any hanging microtasks: part 1', () => {
+  render(<App />)
+  expect(document.body.textContent).toBe('123')
+  expect(log).toEqual([])
+})
+
+it('cleanup-after-each does not leave any hanging microtasks: part 2', () => {
+  expect(log).toEqual([0])
+  expect(document.body.innerHTML).toBe('')
+})
--- a/src/act-compat.js
+++ b/src/act-compat.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
-import {reactDomSixteenPointNineIsReleased} from './react-dom-16.9.0-is-released'
 import * as testUtils from 'react-dom/test-utils'
+import {reactDomSixteenPointNineIsReleased} from './react-dom-16.9.0-is-released'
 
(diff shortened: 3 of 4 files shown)


threepointone:
I wonder how this would work with `resetModules()`. I'll update the PR to be a bit more defensive.

alexkrolick:
> Why not make cleanup() return a promise?: That would be a breaking change, especially for people already using cleanup() in their tests. You could consider it for a next major.

I think we would remove `cleanupAsync` if we changed `cleanup` to return a promise, so that would create double thrash for anyone who switches now.

What if we do the breaking change release now and either document a workaround for the current version or release this as a non-recommended minor version on top of the current release?

threepointone:
> I think we would remove cleanupAsync if we changed cleanup to return a promise, so that would create double thrash for anyone who switches now.

Right, so don't even consider cleanupAsync as a public api right now?

threepointone:
I updated the PR to do a dynamic require() in `cleanup-after-each`

threepointone:
I could even just move the code into cleanup-after-each, and not have a new export at all

alexkrolick:
> Right, so don't even consider cleanupAsync as a public api right now?

I suspect these usages are just as common if not more so than `import "@testing-library/react/cleanup-afterEach" `:

```js
afterEach(cleanup)
```

```js
afterEach(() => {
  // Something else ...
  cleanup()
})
```

kentcdodds:
Yeah, I wish that people would just import the thing, but I suspect many are using cleanup directly 😔 I'll do a twitter poll for curiosity reasons, but I think we should leave this PR as-is, not document `cleanupAsync` and then include it in the next major. I don't see much reason to make a breaking change for this feature.

threepointone:
Alright, I updated the PR so it's only in `cleanup-after-each` now, and we can decouple this PR from whatever the call on `cleanup` is

kentcdodds:
Here it is: https://twitter.com/TestingLib/status/1158795165420572672

src/cleanup-after-each.js
```diff
@@ -0,0 +1,6 @@
+module.exports = async function cleanAfterEach() {
```
kentcdodds: I'd rather this file not exist. Could we just put this code in the `cleanup-after-each.js` file?

threepointone: you probably want this file to be tested / compiled / validated though now.

kentcdodds: I suppose it's fine. The thing is that it's not going to be included in the UMD because it's not imported into the main export, but that's actually not a big deal because I don't think many people use the UMD anyway.

alexkrolick: Kinda weird that is called `cleanup-after-each` when it doesn't do that though 🙁

threepointone: Good point, will rename

threepointone: Done!

kentcdodds:
Thanks Sunil!

PR #430, opened 2019-08-08
feat(cleanup): automatically cleanup if afterEach is detected

### LINKED ISSUE #428: Have render call cleanup

Right now we have to call `cleanup` before each test. This tends to be confusing for beginners especially if they come from other libraries (Enzyme) that do not need this step. It would be good if `render` could call `cleanup` automatically.

### Suggested implementation:

I didn't dig into the code too much but it should be as easy as having `render` call `cleanup` as the first thing it does.

I'm trying to think about a use-case where you might want to run two `render`s in the same test or where you might want to manually run `cleanup`—which would invalidate this idea—but can't come up with anything.

### Teachability, Documentation, Adoption, Migration Strategy:

We would have to remove the documentation about `cleanup`. Potentially we would have to introduce a breaking change.

kentcdodds:
**What**:

automatically call `cleanup` (with async act) if `afterEach` is detected.

**Why**:

Closes #428

This effectively means that 90% of users wont even have to know about `cleanup` at all. Which is a huge improvement.

**How**:

You can disable this with the `RTL_SKIP_CLEANUP` environment variable if you so choose, but it's recommended to have cleanup work this way.

**Checklist**:

      [docs site](https://github.com/alexkrolick/testing-library-docs) Let's get this reviewed/merged first, then we can start working on the docs.
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/auto-cleanup-skip.js
+++ b/src/__tests__/auto-cleanup-skip.js
@@ -1,0 +1,18 @@
+import React from 'react'
+
+let render
+beforeAll(() => {
+  process.env.RTL_SKIP_CLEANUP = 'true'
+  const rtl = require('../')
+  render = rtl.render
+})
+
+// This one verifies that if RTL_SKIP_CLEANUP is set
+// that we DON'T auto-wire up the afterEach for folks
+test('first', () => {
+  render(<div>hi</div>)
+})
+
+test('second', () => {
+  expect(document.body.innerHTML).toEqual('<div><div>hi</div></div>')
+})
--- a/src/__tests__/auto-cleanup-skip.js
+++ b/src/__tests__/auto-cleanup-skip.js
@@ -8,7 +8,7 @@ beforeAll(() => {
 })
 
 // This one verifies that if RTL_SKIP_CLEANUP is set
-// that we DON'T auto-wire up the afterEach for folks
+// then we DON'T auto-wire up the afterEach for folks
 test('first', () => {
   render(<div>hi</div>)
 })
--- a/src/__tests__/auto-cleanup-skip.js
+++ b/src/__tests__/auto-cleanup-skip.js
@@ -2,12 +2,12 @@ import React from 'react'
 
 let render
 beforeAll(() => {
-  process.env.RTL_SKIP_CLEANUP = 'true'
+  process.env.RTL_SKIP_AUTO_CLEANUP = 'true'
   const rtl = require('../')
   render = rtl.render
 })
 
-// This one verifies that if RTL_SKIP_CLEANUP is set
+// This one verifies that if RTL_SKIP_AUTO_CLEANUP is set
 // then we DON'T auto-wire up the afterEach for folks
 test('first', () => {
   render(<div>hi</div>)
--- a/src/index.js
+++ b/src/index.js
@@ -142,6 +142,16 @@ fireEvent.select = (node, init) => {
   fireEvent.keyUp(node, init)
 }
 
+// if we're running in a test runner that supports afterEach
+// then we'll automatically run cleanup afterEach test
+// this ensures that tests run in isolation from each other
+if (typeof afterEach === 'function' && !process.env.RTL_SKIP_CLEANUP) {
+  afterEach(async () => {
+    await asyncAct(async () => {})
+    cleanup()
+  })
+}
+
 // just re-export everything from dom-testing-library
 export * from '@testing-library/dom'
 export {render, cleanup, fireEvent, act}
--- a/src/index.js
+++ b/src/index.js
@@ -1,163 +1,16 @@
-import React from 'react'
-import ReactDOM from 'react-dom'
-import {
-  getQueriesForElement,
-  prettyDOM,
-  fireEvent as dtlFireEvent,
-  configure as configureDTL,
-} from '@testing-library/dom'
-import act, {asyncAct} from './act-compat'
-
-configureDTL({
-  asyncWrapper: async cb => {
-    let result
-    await asyncAct(async () => {
-      result = await cb()
-    })
-    return result
-  },
-})
-
-const mountedContainers = new Set()
-
-function render(
-  ui,
-  {
-    container,
-    baseElement = container,
-    queries,
-    hydrate = false,
-    wrapper: WrapperComponent,
-  } = {},
-) {
-  if (!baseElement) {
-    // default to document.body instead of documentElement to avoid output of potentially-large
-    // head elements (such as JSS style blocks) in debug output
-    baseElement = document.body
-  }
-  if (!container) {
-    container = baseElement.appendChild(document.createElement('div'))
-  }
-
-  // we'll add it to the mounted containers regardless of whether it's actually
-  // added to document.body so the cleanup method works regardless of whether
-  // they're passing us a custom container or not.
-  mountedContainers.add(container)
-
-  const wrapUiIfNeeded = innerElement =>
-    WrapperComponent
-      ? React.createElement(WrapperComponent, null, innerElement)
-      : innerElement
-
-  act(() => {
-    if (hydrate) {
-      ReactDOM.hydrate(wrapUiIfNeeded(ui), container)
-    } else {
-      ReactDOM.render(wrapUiIfNeeded(ui), container)
-    }
-  })
-
-  return {
-    container,
-    baseElement,
-    // eslint-disable-next-line no-console
-    debug: (el = baseElement) => console.log(prettyDOM(el)),
-    unmount: () => ReactDOM.unmountComponentAtNode(container),
-    rerender: rerenderUi => {
-      render(wrapUiIfNeeded(rerenderUi), {container, baseElement})
-      // Intentionally do not return anything to avoid unnecessarily complicating the API.
-      // folks can use all the same utilities we return in the first place that are bound to the container
-    },
-    asFragment: () => {
-      /* istanbul ignore if (jsdom limitation) */
-      if (typeof document.createRange === 'function') {
-        return document
-          .createRange()
-          .createContextualFragment(container.innerHTML)
-      }
-
-      const template = document.createElement('template')
-      template.innerHTML = container.innerHTML
-      return template.content
-    },
-    ...getQueriesForElement(baseElement, queries),
-  }
-}
-
-function cleanup() {
-  mountedContainers.forEach(cleanupAtContainer)
-}
-
-// maybe one day we'll expose this (perhaps even as a utility returned by render).
-// but let's wait until someone asks for it.
-function cleanupAtContainer(container) {
-  ReactDOM.unmountComponentAtNode(container)
-  if (container.parentNode === document.body) {
-    document.body.removeChild(container)
-  }
-  mountedContainers.delete(container)
-}
-
-// react-testing-library's version of fireEvent will call
-// dom-testing-library's version of fireEvent wrapped inside
-// an "act" call so that after all event callbacks have been
-// been called, the resulting useEffect callbacks will also
-// be called.
-function fireEvent(...args) {
-  let returnValue
-  act(() => {
-    returnValue = dtlFireEvent(...args)
-  })
-  return returnValue
-}
-
-Object.keys(dtlFireEvent).forEach(key => {
-  fireEvent[key] = (...args) => {
-    let returnValue
-    act(() => {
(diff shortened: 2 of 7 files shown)


kentcdodds:
cc @threepointone, I'd love your input here.

kentcdodds:
Here's the one issue that I think people will bump up against. If people write their tests like this:

```javascript
const {getByText} = render(<Counter />)
const counterButton = getByText(/^count/i)

test('the counter is initialized to the initialCount', () => {
  expect(counterButton).toHaveTextContent('0')
})

test('when clicked, the counter increments the click', () => {
  fireEvent.click(counterButton)
  expect(counterButton).toHaveTextContent('1')
})
```

Or similar, then they'll be in trouble because we'll auto-cleanup for them and their tests will break in a way that's really confusing.

They should definitely not be writing their tests like that, but I know a lot of people do put each assertion in its own `it`, which is really pointless, but they do it and tests will fail. So we'll need to have a very helpful changelog describing how to use the workaround `RTL_SKIP_CLEANUP` environment variable and why they should fix their tests to be isolated.

kentcdodds:
FWIW, I have this blog post written already: https://kentcdodds.com/blog/test-isolation-with-react

src/__tests__/act.js
```diff
@@ -1,7 +1,5 @@
 import React from 'react'
-import {render, cleanup, fireEvent} from '../'
-
-afterEach(cleanup)
```
kentcdodds: Several of the tests were manually wiring up `cleanup` even though they didn't need to before (and they certainly don't need to now), so I fixed those.

threepointone:
Bit dead today because of the release, so I'll be a while before responding haha. But I don't think you should merge this any time soon, at least not without thinking about it for a day or so. Seems like too little to warrant a major version bump.

src/__tests__/act.js
alexkrolick: There a bunch of docs with `afterEach(cleanup)` too

alexkrolick:
> I know a lot of people do put each assertion in its own `it`

I suspect that most of these people don't have cleanup configured in most of their tests at all 😬, unless they manually do `cleanup` in the `afterAll` (`after`) hook in each `describe()`. Easy to miss and fragile.

bcarroll22:
🤔 I think I'm a fan of this model as a major release. Weighing the tradeoffs personally, this seems like it will save a lot more bugs than it will create. I've seen people get burned multiple times by not importing and running cleanup properly.

If the primary downside is that some test suites will have to be refactored during the upgrade to be better isolated, that seems reasonable.

kentcdodds:
After considering this a lot, I've determined that we should go forward with this change.

The Testing Library family of tools has a primary goal: To enable people to write tests which give them more confidence and are easier to maintain. The kinds of tests that this change could break do not give people more confidence nor are they easy to maintain. And the current requirement to have to wire up cleanup makes using react testing library harder (which is contrary to the goals of the library).

This is worth a breaking change IMO. The "churn" will be a minimum and will only negatively impact people who are not using the library as documented anyway. And for people who have been doing things correctly by wiring up the `afterEach(cleanup)` properly will be able to upgrade and they don't even have to remove their wiring if they don't want to. It will still work either way.

So in a few hours I will be merging this unless there's good reason to not do so.

DaveStein:
Looks like this would be an auto-solve for  https://github.com/testing-library/react-hooks-testing-library/issues/76. I was confused when I looked at your PR, but then saw that cleanup essentially ends up calling `ReactDOM.unmountComponentAtNode(container)`, which is what is exposed simply as `unmount` to public API.

threepointone:
How will this work with concurrent mode (.createRoot()/.unmount())? or do you not factor that in right now?

src/index.js
```diff
@@ -142,6 +142,16 @@ fireEvent.select = (node, init) => {
   fireEvent.keyUp(node, init)
 }
 
+// if we're running in a test runner that supports afterEach
+// then we'll automatically run cleanup afterEach test
+// this ensures that tests run in isolation from each other
+if (typeof afterEach === 'function' && !process.env.RTL_SKIP_CLEANUP) {
+  afterEach(async () => {
+    await asyncAct(async () => {})
```
threepointone: Consider - if ther are any hanging promises after a test is finished, that probably means the user hasn't asserted on the actual stable state of the UI. So this would hide a probable bug.

threepointone: (Maybe this is a similar concern with the async act() in cleanup-after-each, ouch. at least there it's explicit tho.

kentcdodds:
> How will this work with concurrent mode (.createRoot()/.unmount())

I don't know. I guess we don't factor that in right now. Currently you can only `ReactDOM.render` anyway, so what we have today works with what you can do anyway. When we add concurrent mode rendering then perhaps we'll have to adjust cleanup to manage unmounting those nodes as well I guess (if it's different?).

src/index.js
kentcdodds: > if there are any hanging promises after a test is finished, that probably means the user hasn't asserted on the actual stable state of the UI.

I'm not sure I understand how this change impacts that at all. This change is doing exactly the same thing that `cleanup-after-each` is doing. It's just wiring it up automatically.

To be clear: all this change does is make it so people don't have to wire up cleanup-after-each themselves (which they should be doing). If they're doing that already, then this change will make 0 impact on them.

kentcdodds:
To be clear: all this change does is make it so people don't have to wire up cleanup-after-each themselves (which they should be doing). If they're doing that already, then this change will make 0 impact on them.

So **the only people who will experience any bumps in this transition are the people who are using RTL incorrectly.**

threepointone:
If you're going all in on a major update, why do it just for this one feature, and so soon, is my major concern. I think you should think deeply on the implications, and at least bounce around some ideas.

Alternate idea for this 'feature': Custom describe/test helpers? You'd have a lot more freedom there too.

src/index.js
threepointone: Maybe this shouldn't be an act(), but it should flush promises anyway. So if there are any stray async updates, the warning will fire, and the dev will account for it in their test.

threepointone: > I'm not sure I understand how this change impacts that at all. This change is doing exactly the same thing that cleanup-after-each is doing. It's just wiring it up automatically.

I'm suggesting it should be removed from that too.

kentcdodds:
I definitely don't want custom `describe/test` helpers. That increases the API surface/intrusiveness significantly and the number of people who wouldn't use those helpers is huge. We'd basically split the community between people who use those helpers and those who don't. I'm sorry Sunil, but I'm pretty firm on this.

kentcdodds:
Also, I'm not as concerned about publishing major version bumps to my libraries. It's just a number, people using the current version get a great library. People who upgrade have very little work to do to upgrade (less than if we batched a ton of changes for a major version change).

I personally feel like most people wont need to make any changes for this major version change, so upgrading to this version will require very minimal work.

threepointone:
Sure, it was a suggestion. I'm trying to point out that you're set on a major, then you can rethink stuff beyond this one feature. 

If you're set on making this change, go for it, I'll keep my feedback to myself.

kentcdodds:
> you're set on a major, then you can rethink stuff beyond this one feature.

That's a good point.

I'm definitely open to suggestions and feedback. I just thought I had already explained my thoughts around your feedback about the custom describe/test helpers.

src/__tests__/act.js
kentcdodds: I made these unrelated polish changes to master so this PR could be slimmer.

kentcdodds:
There's one other breaking change that we could potentially include with this one: https://github.com/testing-library/dom-testing-library/issues/314

I'm just worried that this will be a significant change and there's too much inertia behind the `debug` function that it would be confusing for people 😬 But I could be wrong. I'll make a PR on DTL right now to see what that would be like.

benlesh:
+1 for major version bumps for anything that might even be remotely breaking.

(I'll go back to lurking now)

bcarroll22:
@threepointone, I see where you're coming from. I think allowing semantic release to do versions in the TL family has mitigated anxiety/hesitancy toward major releases for us. And that's been around since the beginning. Since every merged PR is pretty much a new release/version automatically, as long as we're honest about when things break, get added, or get fixed, everything stays pretty transparent and our users are used to it (at least I don't recall any explicit complaints about our versioning).

Back to this PR though, I really think this is great. 

* Cleaning up behind the scenes will be so much simpler for those who do it explicitly today
* Adoption for new users is so much better this way because cleanup won't be forgotten
* Those who have failing tests because of this can use .env to opt out until they can refactor

I don't see any concern with the feature, but I do see where Sunil is coming from with the major version bump. That said, given the way we have always versioned releases, I honestly think it'll be okay for our users.

kentcdodds:
As an update, https://github.com/testing-library/dom-testing-library/pull/336 is going to be a small breaking change, so we'll release that and then release this.

Will still wait on a little more time to give folks time to comment on both PRs.

src/__tests__/auto-cleanup-skip.js
```diff
@@ -0,0 +1,18 @@
...
+
+let render
+beforeAll(() => {
+  process.env.RTL_SKIP_CLEANUP = 'true'
+  const rtl = require('../')
+  render = rtl.render
+})
+
+// This one verifies that if RTL_SKIP_CLEANUP is set
+// that we DON'T auto-wire up the afterEach for folks
```
afontcu:
```suggestion
// then we DON'T auto-wire up the afterEach for folks
```

src/index.js
```diff
@@ -142,6 +142,16 @@ fireEvent.select = (node, init) => {
   fireEvent.keyUp(node, init)
 }
 
+// if we're running in a test runner that supports afterEach
+// then we'll automatically run cleanup afterEach test
+// this ensures that tests run in isolation from each other
+if (typeof afterEach === 'function' && !process.env.RTL_SKIP_CLEANUP) {
```
eps1lon: We should guard against process and process.env not being defined for people running their tests in actual browsers.

kentcdodds:
> I'm not a big fan of having side-effects in modules

I feel the same way. I'm fine with the `pure` approach. I'll refactor this to that.

kentcdodds:
Actually, I think we'll support both. Because lots of people who would need to make this migration would probably prefer the `env` approach.

kentcdodds:
Ok, both mechanisms are supported. So people who don't want this behavior can either set `RTL_SKIP_AUTO_CLEANUP` to `'true'` (or anything for that matter) OR they can import from `@testing-library/react/pure`.

But 99% of people wont even know this is a thing (or want it) and that's a huge win.

kentcdodds:
Basically, we're changing the default for the more common case. Before, 99% of people had to wire up the cleanup and 1% didn't bother (and their tests were probably not very good). Now, 99% of people don't have to do anything, 0.9% of people will have to figure out a better way to write tests, and 0.1% of people will have to figure out how to skip out of the auto-cleanup stuff.

Those numbers are made up. It's probably even more favorable to be honest.

kentcdodds:
Ok, this is going to go in the next few minutes.

kentcdodds:
I'm going to upgrade @testing-library/dom with this PR as well. So we'll get a single release for both breaking changes (because we re-export everything from @testing-library/dom).

kentcdodds:
Here it goooooes!

reyronald:
Wow this is great Kent, thank you so much! Totally in favor of this change.

Just wanted to chime in to say that I think the affected users will be higher than 1% or 0.1%. I do contractor work so I see many different codebases from different teams and companies often and the style you described in https://github.com/testing-library/react-testing-library/pull/430#issuecomment-519702557 is far more common than one would think. I suspect it might be because of the micro-optimization tendencies and that low-performance phobia that some developers have that they think that re-using the container/wrapper element across several tests will make their test suites faster. Or maybe they don't feel comfortable rendering the components each time on each tests because it doesn't feel DRY ?

Regardless, I will not make up any numbers or statistics, and like I said I totally agree with the change, just wanted to provide a different perspective and expectations based on my personal anecdotal experience, just because I can think of many repos that I've worked with this year alone that I know will have test failures.

kentcdodds:
You're probably right @reyronald. Hopefully this change will help people rethink their testing practices.

That said, I decided to make disabling this behavior as easy as enabling the original behavior was: https://github.com/testing-library/react-testing-library/pull/435

PR #433, opened 2019-08-09
fix(cleanup-after-each): remove cleanup-after-each file.

### LINKED ISSUE #432: Remove cleanup-after-each
This is now automatic and shouldn't exist anymore 😅

kentcdodds:
**What**: remove cleanup-after-each file.

**Why**: This is automatic now, so there is no reason to have it

**How**: Remove all the related files and the entry in package.json `files` array.

**Checklist**:

      [docs site](https://github.com/alexkrolick/testing-library-docs) - Will do this soon
- [x] Tests
- [x] Ready to be merged
      

Closes #432


FredyC:
I think you went too fast with that, could have been at least deprecated to be removed in next major. It's a relatively small thing but adds unnecessary overhead to upgrading.

FredyC:
Deprecation warning would also teach people about the feature in case someone is not reading release notes too carefully.

kentcdodds:
You're right. I'll add it back with a deprecation warning.

kentcdodds:
Publishing the fix now. I was hasty. Thanks for calling me out.

PR #435, opened 2019-08-10
feat: add dont-cleanup-after-each.js

kentcdodds:
**What**: add dont-cleanup-after-each.js

**Why**:

This makes it easier for folks who don't want to write their tests in an isolated way. It's totally not recommended, but it will hopefully reduce pain while people migrate from one testing style to another.

**How**:

This is just a simple file that sets the RTL_SKIP_AUTO_CLEANUP environment variable so that RTL doesn't setup automatic cleanup. 

**Checklist**:

      [docs site](https://github.com/alexkrolick/testing-library-docs) [done](https://github.com/testing-library/testing-library-docs/pull/215)
- [ ] Tests N/A
- [x] Ready to be merged
      

I decided to do this after reading https://github.com/testing-library/react-testing-library/pull/430#issuecomment-520142114

### DIFF
--- a/dont-cleanup-after-each.js
+++ b/dont-cleanup-after-each.js
@@ -1,0 +1,1 @@
+process.env.RTL_SKIP_AUTO_CLEANUP = true
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "dist",
     "typings",
     "cleanup-after-each.js",
+    "dont-cleanup-after-each.js",
     "pure.js"
   ],
   "keywords": [


kentcdodds:
I'd love someone else to merge this. I've been a bit merge-happy recently 😅

dont-cleanup-after-each.js
```diff
@@ -0,0 +1 @@
+process.env.RTL_SKIP_AUTO_CLEANUP = true
```
alexkrolick: Does this need to be browser-safe?

kentcdodds: I don't think it is worth the effort.

Issue #436, opened 2019-08-11
typescript declaration of `act` should match React 16.9.x


davidje13:
Since 16.9.0 was released with a fix for https://github.com/facebook/react/issues/14769, `act` returns a `Promise` (and optionally takes in a `Promise`).

[This line](https://github.com/testing-library/react-testing-library/blob/master/typings/index.d.ts#L46) should be updated to allow this, or better yet, it should forward the type `typeof act` from `@types/react-dom` ([see here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-dom/test-utils/index.d.ts#L295), though I'm not sure if that's 100% correct since I think it is now allowed to pass in a promise).

Right now the workaround is to import `act` directly from `react-dom` (since `act` here is just a pass-through), or to define:

```ts
function asyncAct(fn: () => Promise<any> | void): Promise<void> {
  return act(fn) as any as Promise<void>;
}
```

PR #437, opened 2019-08-11
chore(typescript): remove typings

alexkrolick:
See #436, https://github.com/testing-library/dom-testing-library/pull/337

Remove the TypeScript typedefs from this repo in favor of having the community maintain them at DefinitelyTyped.

**Why**

There are no core teammembers with the TS knowledge needed to maintain the types, and they are not well tested or integrated with the codebase.

---

See also https://github.com/reduxjs/redux/issues/3500


nickmccurdy:
It would be best to wait to merge until the types are published in DefinitelyTyped.

eps1lon:
> There are no core teammembers with the TS knowledge needed to maintain the types, and they are not well tested or integrated with the codebase.

This seems a bit too eager I feel. The issue is open for 9 hours, you opened this one 2 hours after the original issue was submitted. If the expectation for members with TS knowledge is to fix them within 2 hours then yes there are none.

Other than that I'm happy to fix #436. It might even be possible so that we don't need to maintain `act` typings.

afontcu:
As far as I can tell, even Microsoft suggest to use DefinitelyTyped if the project isn't written in TypeScript: 

> If your package is not written in TypeScript then the second is the preferred approach. (that is, publishing to the @types organization on npm).

([source](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html))

eps1lon:
I would always recommend this as well. But the way this should be done is by creating definitions in `@types` first before removing those here i.e. offer a migration path for a breaking change.

FredyC:
I want to point that moving to DefinitelyTypes doesn't automatically ensure up-to-date declarations. Unless someone from the community can step up, it will become stale maybe even faster than having it inside the repo. Hopefully, RTL is popular enough for that not to happen.

From a user point of view, it's much easier having types in the package so they don't need to be installed separately.

What about having a `index.d.ts` in a package that will reference `@types/*` and add those to dependencies? It's such a small dependency it will hardly matter (especially for dev tool) and it gives the best of two worlds.

nickmccurdy:
> I want to point that moving to DefinitelyTypes doesn't automatically ensure up-to-date declarations. Unless someone from the community can step up, it will become stale maybe even faster than having it inside the repo. Hopefully, RTL is popular enough for that not to happen.

No maintenance is automatic, but putting types in the same package means that maintainers are responsible for reviewing TypeScript pull requests that they may not understand, when there's already a large community of TypeScript developers at DefinitelyTyped that help with reviewing pull requests.

> From a user point of view, it's much easier having types in the package so they don't need to be installed separately.

Having types in the same package is much worse for users. If there's a breaking change in the types, you either don't update the major version in which case TypeScript users get an unpleasant surprise (happens more frequently when package maintainers don't use TypeScript themselves), or you do in which case the version keeps changing for TS and JS users when it only affects TS users. Version management is much easier when types are in a separate package (except for when the package is written in TS, which keeps the types in sync).

> What about having a `index.d.ts` in a package that will reexport `@types/*` and add those to dependencies? It's such a small dependency it will hardly matter (especially for dev tool) and it gives the best of two worlds.

Are you suggesting that `@types/testing-library__react` depends on the other `@types/testing-library__*` packages? This is a good idea, and is recommended by the TypeScript team.

weyert:
I am using TypeScript at work and happy to keep them up to date a:)
My colleague is part of DefinetlyTyoed so should be able to get PR reasonable quickly merged :)

alexkrolick:
> > What about having a index.d.ts in a package that will reexport @types/* and add those to dependencies? It's such a small dependency it will hardly matter (especially for dev tool) and it gives the best of two worlds.

> Are you suggesting that @types/testing-library__react depends on the other @types/testing-library__* packages? This is a good idea, and is recommended by the TypeScript team.

I think @FredyC meant `@testing-library/react` could have `@types/testing-library__react` as a dependency and re-export it. I am not sure if that is recommended, it's the first time I'm hearing of that approach.

@weyert would you be willing to start a PR to DT to add the types? As others have pointed out the responsible approach is to get those ready first to ensure a migration path.

weyert:
Sure, I will have a look st it tomorrow. Earlier I already made one for user-event :)

FredyC:
@alexkrolick 
> I think @FredyC meant `@testing-library/react` could have `@types/testing-library__react` as a dependency and re-export it.

You got that correctly. I never heard of it either, but it just occurred to me it could solve both pains at once.

@nickmccurdy 
> Having types in the same package is much worse for users.

If I understand you correctly, you mean this in case the typings are out of date? Sure, that's a problem. I meant it more generally, be it that project is written in TypeScript or not, with up to date typings it's better DX than having to check if `@types` is available and install it.

weyert:
I can imagine that inherited types of dom package in the React type definitions would be possible 

Weyert

FredyC:
> I can imagine that inherited types of dom package in the React type definitions would be possible

Sure, that is one part, but I am talking about having `index.d.ts` in this package that reexports `@types/testing-library__react`. That way user doesn't need to install types and they will be maintained over at DT.

afontcu:
I see that several Testing Library projects have types-related debates open (RTL, [DTL](https://github.com/testing-library/dom-testing-library/pull/337), [VTL](https://github.com/testing-library/vue-testing-library/pull/74), [user-event](https://github.com/testing-library/user-event/issues/140)...). I wonder if there's a way of avoiding repetition and to provide the same experience to our users regardless of the Testing Library flavor or utility they're using.

Since DTL is the foundation of many packages, I guess RTL and VTL types would extend from DTL's and customize them accordingly. Am I right?

Does it make sense? Is this the direction we are going? 🙌 Maybe the debate should be unified too, so all efforts are focused?

weyert:
> Does it make sense? Is this the direction we are going? 🙌 Maybe the debate should be unified too, so all efforts are focused?

Yeah, that's the idea I am having for now. I think that would be the most convenient way to get the query type defs into the projects?

nickmccurdy:
> @nickmccurdy
> 
> > Having types in the same package is much worse for users.
> 
> If I understand you correctly, you mean this in case the typings are out of date? Sure, that's a problem. I meant it more generally, be it that project is written in TypeScript or not, with up to date typings it's better DX than having to check if `@types` is available and install it.

You're right that the types themselves are going to have problems whether or not they're in the same package, but my point is that version management is difficult to do well without a separate types package. This is why TypeScript recommends removing types from packages not written in TypeScript.

kentcdodds:
For the record. I'm in favor of these changes. I think it'll be better for everyone. Thanks for working on it @weyert!

cimbul:
@weyert, are you also planning on handling the DefinitelyTyped PR for dom-testing-library? I've got a DT PR ready for VTL (see discussion on testing-library/vue-testing-library#74), but I'm waiting on DTL to be moved there first. If you were only going to handle RTL, I can take a pass at it, but I wasn't clear.

weyert:
I am happy to take it over :)

weyert:
Sorry, didn’t had time yet to work on it. I will make sure it happens this week. I will try before Friday otherwise it will be Friday.

Got lost in the world of two job schedulers talking to the same database today! Five hours gone. If you use Hangfire check the heartbeats and the hostname 😀💪😴

MichaelDeBoey:
The redux repo is more leaning towards rewriting the code in TS.

Is this something we want to consider too for the testing-library packages?
Maybe only DTL (since redux is considered, but redux-thunk, reselect, ... aren’t at this moment)? 🤔

kentcdodds:
I've held back from writing my OSS in TypeScript because I don't want to raise the barrier for contributors. So no, I don't think that's what we're going to do.

alexkrolick:
Started on a draft PR just to see what using DefinitelyTyped is like https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37606

First thing is the stylistic checkers definitely don't like some of the existing type names & patterns...

<details>
  <summary>Errors</summary>

```

Error in testing-library__dom
Error: ~/DefinitelyTyped/types/testing-library__dom/config.d.ts:1:18
ERROR: 1:18   interface-name                   interface name must not have an "I" prefix
ERROR: 3:23   ban-types                        Don't use 'Function' as a type. Avoid using the `Function` type. Prefer a specific function type, like `() => void`.
ERROR: 3:42   no-unnecessary-generics          Type parameter T is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md
ERROR: 6:18   interface-name                   interface name must not have an "I" prefix

~/DefinitelyTyped/types/testing-library__dom/get-queries-for-element.d.ts:16:11
ERROR: 16:11  strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
ERROR: 26:11  strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md

~/DefinitelyTyped/types/testing-library__dom/matches.d.ts:27:8
ERROR: 27:8   strict-export-declare-modifiers  'declare' keyword is redundant here. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md

~/DefinitelyTyped/types/testing-library__dom/query-helpers.d.ts:23:14
ERROR: 23:14  prefer-declare-function          Use a function declaration instead of a variable of function type. See: https://github.com/Microsoft/dtslint/blob/master/docs/prefer-declare-function.md
ERROR: 24:14  prefer-declare-function          Use a function declaration instead of a variable of function type. See: https://github.com/Microsoft/dtslint/blob/master/docs/prefer-declare-function.md
ERROR: 32:6   strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
ERROR: 62:20  expect                           Compile error in typescript@3.2 but not in typescript@3.3.
Fix with a comment '// TypeScript Version: 3.3' just under the header.
Trailing comma not allowed.
ERROR: 64:14  prefer-declare-function          Use a function declaration instead of a variable of function type. See: https://github.com/Microsoft/dtslint/blob/master/docs/prefer-declare-function.md

~/DefinitelyTyped/types/testing-library__dom/testing-library__dom-tests.ts:1:25
ERROR: 1:25   expect                           TypeScript@next compile error:
Cannot find module '@testing-library/dom'.

~/DefinitelyTyped/types/testing-library__dom/wait-for-dom-change.d.ts:5:13
ERROR: 5:13   no-unnecessary-generics          Type parameter T is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md

    at ~/DefinitelyTyped/node_modules/dtslint/bin/index.js:190:19
    at Generator.next (<anonymous>)
    at fulfilled (~/DefinitelyTyped/node_modules/dtslint/bin/index.js:5:58)


Error in testing-library__dom
Error: ~/DefinitelyTyped/types/testing-library__dom/config.d.ts:1:18
ERROR: 1:18   interface-name                   interface name must not have an "I" prefix
ERROR: 3:23   ban-types                        Don't use 'Function' as a type. Avoid using the `Function` type. Prefer a specific function type, like `() => void`.
ERROR: 3:42   no-unnecessary-generics          Type parameter T is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md
ERROR: 6:18   interface-name                   interface name must not have an "I" prefix

~/DefinitelyTyped/types/testing-library__dom/get-queries-for-element.d.ts:16:11
ERROR: 16:11  strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
ERROR: 26:11  strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md

~/DefinitelyTyped/types/testing-library__dom/matches.d.ts:27:8
ERROR: 27:8   strict-export-declare-modifiers  'declare' keyword is redundant here. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md

~/DefinitelyTyped/types/testing-library__dom/query-helpers.d.ts:23:14
ERROR: 23:14  prefer-declare-function          Use a function declaration instead of a variable of function type. See: https://github.com/Microsoft/dtslint/blob/master/docs/prefer-declare-function.md
ERROR: 24:14  prefer-declare-function          Use a function declaration instead of a variable of function type. See: https://github.com/Microsoft/dtslint/blob/master/docs/prefer-declare-function.md
ERROR: 32:6   strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
ERROR: 62:20  expect                           Compile error in typescript@3.2 but not in typescript@3.3.
Fix with a comment '// TypeScript Version: 3.3' just under the header.
Trailing comma not allowed.
ERROR: 64:14  prefer-declare-function          Use a function declaration instead of a variable of function type. See: https://github.com/Microsoft/dtslint/blob/master/docs/prefer-declare-function.md

~/DefinitelyTyped/types/testing-library__dom/testing-library__dom-tests.ts:1:25
ERROR: 1:25   expect                           TypeScript@next compile error:
Cannot find module '@testing-library/dom'.

~/DefinitelyTyped/types/testing-library__dom/wait-for-dom-change.d.ts:5:13
ERROR: 5:13   no-unnecessary-generics          Type parameter T is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md

```

</details>

eps1lon:
@alexkrolick I'll help you out. I've contributed types there before.

weyert:
🤷‍♀️

weyert:
Clearly my help is not needed.

eps1lon:
> Clearly my help is not needed.

Sorry I got kind of got lost who is doing what. But an additional set of eyes to review is always appreciated: DefinitelyTyped/DefinitelyTyped#37606 for the original work and https://github.com/alexkrolick/DefinitelyTyped/pull/1 for a green build.

alexkrolick:
@eps1lon thanks for https://github.com/alexkrolick/DefinitelyTyped/pull/1

@weyert sorry for stepping on your toes for this batch of changes.

If RTL and DTL are working (pending 👀), @weyert could you take on user-event and vue? The more we can get ported over in one go the less churn there is in the ecosystem.

kentcdodds:
I'd like to discuss whether to make this a breaking change. Technically any change to the typings could be considered a breaking change, but we've always treated them as fixes. I would like to do that for this change as well, but I'm open to hear what others have to say about this.

weyert:
Sorry, I have been moody today. One of those days. I think we haven't change user-event since you merged my PR which was shared on DefinitelyTyped?

weyert:
@alexkrolick Updated the type defs: https://github.com/alexkrolick/DefinitelyTyped/pull/2

johnnyreilly:
Definitely Typed maintainer here - the linked PR is marked as draft right now. Let me know when it's ready and I'll take a look.

weyert:
@johnnyreilly I have made a new PR in @alexkrolick which adds my name and updates user-event. 
@alexkrolick Feel free to mark your PR as ready after my PR is merged. I will make a new PR this week for the Vue typedefs.

cimbul:
@weyert I've already written the testing-library-vue types. They're just waiting on the testing-library-dom DefinitelyTyped PR to be merged.

alexkrolick:
> I've already written the testing-library-vue types

👍 Great, maybe we can get to Cypress and jest-dom as well if @weyert is up for it

PR is ready for review I think https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37606

One question: is this is the right naming convention? `@types/testing-library__react`, `@types/testing-library__dom`

> I'd like to discuss whether to make this a breaking change. 

It is. That may mean that we need to immediately copy the types in the PR to the "v9" (current RTL version) subfolder and make a new version for the latest (v10) version. Not sure if TS will validate the match otherwise.

johnnyreilly:
I've approved the DT PR. I can merge it if you'd like now.

As an aside, I think it's an unusual choice to move type definitions out of a package where it's maintained alongside the library itself.  I think it may be the first time it's happened! There's a whole process around types leaving Definitely Typed to be maintained with their packages. All that said, I'm happy to support you in your choice. Let me know when you'd like me to hit the button 😁

I'd advise that you release a new version of `react-testing-library` as soon as possible *without* TypeScript definitions in. When that is out people can start consuming the `@types/react-testing-library` package which should by then have been published to npm. (Last time I checked it takes around 4 hours to get published)

alexkrolick:
> You might as well mark the type definitions now with the new major version number since it won't ever be applicable for the current version number.

👍 bumped the versions in the PR

> When that is out people can start consuming the @types/react-testing-library package

The npm package names are now scoped: `@testing-library/react`. Is this the right convention for translating to `@types` packages? 👉 `@types/testing-library__react`

johnnyreilly:
> The npm package names are now scoped: `@testing-library/react`. Is this the right convention for translating to `@types| packages? 👉 `@types/testing-library__react`

That's certainly my understanding yes. See: https://stackoverflow.com/questions/47296731/how-can-i-install-typescript-declarations-for-scoped-namespaced-packages-via-ty

> Types for a scoped package `@foo/bar` should go in `types/foo__bar`. Note the double underscore.

kentcdodds:
> What about having a index.d.ts in a package that will reexport @types/* and add those to dependencies? It's such a small dependency it will hardly matter (especially for dev tool) and it gives the best of two worlds.

I'd like to explore this idea again. What would be the downside to doing this?

johnnyreilly:
> I'd like to explore this idea again. What would be the downside to doing this?

Hey @kentcdodds!

I've never encountered this being done before. On the face of it I'm not sure there would be a downside. I don't know if TypeScript would get confused by `testing-library` both transitively installing `@types/testing-library__react` as well as exporting it locally.

It may be fine, in which case no downside. Needs experimentation to confirm.

Also I would say there's some upside with this approach as the  `@types/testing-library__react` convention for scoped packages in not intuitive and generally not known. Makes discoverability poor.

alexkrolick:
> library__react/index.d.ts:1:1
ERROR: 1:1  npm-naming  The types for testing-library__react must match a version that exists on npm.
You should copy the major and minor version from the package on npm.

Should go away after publish, if that's needed. Else the last commit can be reverted.

weyert:
Cool,do you want me to work on the type definitions for the Vue packs for in DefinitelyTyped?

johnnyreilly:
Nice. I think we're waiting for a steer from @kentcdodds about the direction he'd like this to go in. I stand ready to support when that's clear.

kentcdodds:
I just woke up.

If we re-export the types, then this shouldn't be a breaking change right? Let's give that a try and see how it works out.

johnnyreilly:
Okay cool - let's get the version number changed back in the typings and get it merged to Definitely Typed. Then we can advance this PR.

alexkrolick:
version bump reverted

alexkrolick:
> Cool, do you want me to work on the type definitions for the Vue packs for in DefinitelyTyped?

Once DTL is ready, let's see if re-exporting works, and if it does, move VTL, user-event, jest-dom, etc. to DT.

weyert:
Sounds good to me! Thanks for version change. I have gotten lost with work glitches.

cimbul:
> Once DTL is ready, let's see if re-exporting works, and if it does, move VTL, user-event, jest-dom, etc. to DT. (Might still do it if re-exporting doesn't work).

@alexkrolick @weyert Again folks, don't waste your time on vue-testing-library. It's done. I'm just waiting for dom-testing-library to get to DefinitelyTyped so I don't have to add it as an external dependency.

As a side note, whether or not you add `@types/testing-library__dom` as a dependency of `@testing-library/dom` has zero impact on getting testing-library-vue into DefinitelyTyped. The `@types/testing-library__vue` package will depend on `@types/testing-library__dom`, not `@testing-library/vue`. It's also unaffected by the versioning question, because testing-library-vue never had types in the main repo to begin with.

kentcdodds:
Perfect @cimbul. Thank you!

It looks like this was published 20 minutes ago! https://www.npmjs.com/package/@types/testing-library__dom
https://www.npmjs.com/package/@types/testing-library__react

So we should be good to update this PR to add those as deps and re-export them. Then we should be good to merge I think.

nickmccurdy:
Is the plan to make the types package a dependency of the official testing library packages? I wouldn't recommend that, it makes it harder to update the types when they're updated independently from the source package. You'll get some of the advantages of DefinitelyTyped for developing types, but you can't fully use the benefits of separate versioning without completely separating the packages. Of course, this would be a breaking change for TypeScript users, but I think it's worth it.

kentcdodds:
The way I see it, you have two types of users:

1. Those who prefer to install the typings packages.
2. Those who don't know about them and/or don't care too much/only benefit when vscode picks the type defs up for them.

So if we re-export, then group 1 can continue to install the typings packages and get the latest version without trouble. It makes no difference for them that the typings are included (AFAIK). And group 2 will get the typings automatically. They may fall out of date, but they're still better off than not having the typings at all.

I wonder if instead of re-exporting, we just added them as a dependency, but didn't re-export them. That way they're at least installed so group 2 can get the benefit. 🤔

kentcdodds:
I say, why don't we just try this and see what happens. If it doesn't work well, then we can push the major version bump. Unless someone wants to do a more robust test locally to see what happens.

johnnyreilly:
> I say, why don't we just try this and see what happens. If it doesn't work well, then we can push the major version bump. 

Bias to action. I like it.

kentcdodds:
I have a few minutes to work on this right now, so I'll make it happen for both DTL and then RTL and then we can see what happens...

kentcdodds:
Alrighty. I did a real test and it's looking very promising. We wont need to re-export after all :)

https://github.com/testing-library/dom-testing-library/pull/337/files#r314454413

nickmccurdy:
I see your point in theory, but I realized that this package uses peerDependencies which could cause issues. We have two options for dealing with the types of the peerDependencies (react and react-dom) which both have their own issues:

**Move type packages of peerDependencies into peerDependencies:** Allows flexible version management and deduping of types dependencies, but requires non-TypeScript users to manually install `@types/react` and `@types/react-dom`. I think this is the best and safest option (besides removing the types dependency entirely) despite being annoying for non-TypeScript users.

**Depend on peerDependency dependencies directly:** This is likely to cause versioning and type conflicts with TypeScript. TypeScript has similar issues to frameworks like React when there are multiple versions of the same package. If you have two types package for the same package installed, you will get tons of type errors and will not be able to compile unless you use a compatible version that can dedupe with the downstream package's type dependencies.

kentcdodds:
Actually, there's another option and [I just learned about this approach recently](https://github.com/jest-community/jest-runner-eslint/pull/70). The idea is that you make your peerDependency version range as wide as possible, and then you make your devDependency version range equally as wide. That way if someone has installed it themselves, then it will de-dupe to whatever they have installed, and if they don't, then it will just install the latest version for the given range.

I think this will work really well for our scenario.

As it is, the typings have a `*` range for `@types/react` and `@types/react-dom` which is probably not great, but it works fine for our use cases: https://www.unpkg.com/browse/@types/testing-library__react@9.1.0/package.json

So I think we'll be good. And if we're not, we can adjust. But I'm feeling 95% confident that this is going to work nicely for everyone.

nickmccurdy:
I might be misunderstanding, but devDependencies aren't installed when packages are installed via npm/yarn

kentcdodds:
Whoops, meant `dependencies` 😅

kentcdodds:
I'll do a little test by publishing an alpha version :)

kentcdodds:
Here's what it looks like today (with 9.1.1):

image

And here's what it will look like with my changes (with 9.1.2-alpha.0):

image

It's the same 👍 I think we're good!

CONTRIBUTING.md
```diff
@@ -26,45 +26,16 @@ series [How to Contribute to an Open Source Project on GitHub][egghead]
 > can make all of your pull request branches based on this `master` branch.
 > Whenever you want to update your version of `master`, do a regular `git pull`.
 
-## Add yourself as a contributor
```
kentcdodds: I know that cleaning up random things in an unrelated PR is poor practice, but I couldn't help myself 😅

package.json
```diff
@@ -45,13 +43,11 @@
   "dependencies": {
     "@babel/runtime": "^7.5.5",
     "@testing-library/dom": "^6.0.0",
-    "@types/react-dom": "*"
+    "@types/testing-library__react": "^9.1.0"
```
kentcdodds: We don't need the react-dom typings as a dep because they'll come along with testing-library__react :)

kentcdodds:
As soon as the build passes, I'll merge and it'll publish :)

CONTRIBUTING.md
kentcdodds: To be clear, this is managed via a bot... Which I'm about to use :)

kentcdodds:
@all-contributors please add @nickmccurdy for review

kentcdodds:
@all-contributors please add @nickmccurdy for review

kentcdodds:
@all-contributors please add @eps1lon for review

kentcdodds:
@all-contributors please add @weyert for review

kentcdodds:
@all-contributors please add @afontcu for review

kentcdodds:
@all-contributors please add @johnnyreilly for review

kentcdodds:
@all-contributors please add @MichaelDeBoey for review

kentcdodds:
@all-contributors please add @cimbul for review

kentcdodds:
@all-contributors please add @FredyC for review

kentcdodds:
Phew 😌 I think that's everyone. Thanks a TON for all the help here :) I really think this'll be better for everyone!

alexkrolick:
Great work everyone! 🚀

mmiszy:
Since this is a breaking change, shouldn't it be released as 10.0.0?

FredyC:
@mmiszy It's not breaking change because DT types are as a dependency of RTL and (DTL)

mmiszy:
@FredyC does it mean that types are kind of reexported and just work? Thanks for clearing this up :)

johnnyreilly:
I've written this PR up as a blog post that I'll probably publish over the next couple of days. Unless anyone objects I'm going to quote a few of the comments and screenshots in here. Do let me know if there's any objections. I'll share the link here when I publish the blog. I think there's some useful lessons for the community in here around type maintenance / distribution

kentcdodds:
I was thinking about writing it myself. I'm glad you're doing it so I don't have to!

If you could make sure that our motivations are clear that would be great:

We were getting a lot of drive-by contributions to the TypeScript typings and many pull requests would either sit without being reviewed by someone who knows TypeScript well enough, or be merged by a maintainer who just hoped the contributor knew what they were doing. This resulted in a poor experience for TypeScript users who could experience type definition churn and delays, and it became a burden on project maintainers as well (most of us don't know TypeScript very well). Moving the type definitions to DefinitelyTyped puts the maintenance in much more capable hands. And by adding the type definitions to the `dependencies` of React Testing Library, the experience for users is completely unchanged. So it's a huge improvement for the maintenance of the type definitions without any breaking changes for the users of those definitions.

I look forward to reading your blog post!

johnnyreilly:
Awesome - thanks Kent! 

Yes - I'll be sure to make clear the motivations behind this. What I've written roughly mirrors what you've said. I'll probably quote you more directly as I think you've summed it up well.

johnnyreilly:
Blogged! https://blog.johnnyreilly.com/2019/08/symbiotic-definitely-typed.html

Let me know if there's any corrections / clarifications you'd like.  Thanks for your help!

weyert:
Wow great! Thanks for taking the time to write the blog article :)

FredyC:
@johnnyreilly Great post. I don't want to nitpick or collect medals, but I believe it was actually my idea (https://github.com/testing-library/react-testing-library/pull/437#issuecomment-520496015) to add DT types as dependency ;) Unless Kent has done that somewhere else.

kentcdodds:
Almost. Your idea was to re-export them. My idea was to just add them as a dependency so they're installed, but not bother with a re-export. But we're splitting hairs. I do think everyone should be credited for the discussion here.

FredyC:
@kentcdodds Yea, I just wasn't sure it will work by only installing that dep, so I went with a path I know :) You have polished it.

johnnyreilly:
I'll update the article to reflect @FredyC.  I wouldn't like you to feel unrecognised 😄

oudomskn (commented):
Bill

PR #438, opened 2019-08-12
fix: Act type

### LINKED ISSUE #436: typescript declaration of `act` should match React 16.9.x

eps1lon:
**What**:

Closes #436 by putting maintenance burden for `act` types on `@types/react-dom`. async `act` will be supported once https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37554 is merged.

**Why**:

If act is exported from `react-dom/test-utils` we export it as well. The types should reflect that.

**How**:

Conditional types

**Checklist**:

- ~[ ] Documentation added to the
      [docs site](https://github.com/alexkrolick/testing-library-docs)~
- [x] Tests: tested locally by linking package. types from `react-dom/test-utils` are picked up properly.
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.5.5",
-    "@testing-library/dom": "^6.0.0"
+    "@testing-library/dom": "^6.0.0",
+    "@types/react-dom": "*"
   },
   "devDependencies": {
     "@reach/router": "^1.2.1",
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,4 +1,5 @@
 import {queries, Queries, BoundFunction} from '@testing-library/dom'
+import {act as reactAct} from 'react-dom/test-utils'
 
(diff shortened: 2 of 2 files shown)


kentcdodds:
Thanks @eps1lon!

Issue #156, opened 2018-08-10
`onChange` not called for checkboxes in rtl@5


sgoll:
- `react-testing-library` version: 5.0.0
- `react` version: 16.4.2
- `node` version: 8.11.3
- `npm` (or `yarn`) version: 1.9.4

### Relevant code or config:

```tsx
const onChange = jest.fn();
const { container } = render(<input type="checkbox" onChange={onChange} />);
const checkbox = container.firstChild as HTMLInputElement;

// This was working for react-testing-library@4:
// checkbox.checked = true;
// fireEvent.change(checkbox);

// This does not call the event handler:
fireEvent.change(checkbox, { target: { checked: true } });
// This does not call it either:
fireEvent.change(checkbox, { target: { value: true } });

expect(onChange).toHaveBeenCalledTimes(1);
```

### What you did:

Simulate `onChange` event on checkbox.

### What happened:

`onChange` does not get called anymore after upgrade to react-testing-library 5.0.0 and adjusting event handler code to match new shape as outlined in the release notes.

### Reproduction:

Minimal reproduction repository: https://github.com/sgoll/rtl-checkbox

### Problem description:

Testing checkboxes becomes impossible until the event handler get fired/simulated again.

sgoll:
I was able to locate the problem: to improve compatibility with IE8, `react-dom` internally uses the `click` event for checkboxes (and radio inputs)—although mapping it to the `onChange` handler passed as prop.

https://github.com/facebook/react/blob/72434a7686035b4af766ee7d06c070d7f5d6a5f2/packages/react-dom/src/events/ChangeEventPlugin.js#L210

So the proper way to simulate the toggling of a checkbox is as follows:

```ts
const onChange = jest.fn();
const { container } = render(<input type="checkbox" onChange={onChange} />);
const checkbox = container.firstChild as HTMLInputElement;

// This works for react-testing-library@5:
fireEvent.click(checkbox);

expect(onChange).toHaveBeenCalledTimes(1);
```

The change in commit 32f3e65 actually only exposes this behavior by avoiding the simulated call of the `change` handler. Interestingly, running the code above (with `click`) in react-testing-library 4.x.x calls the `onChange` handler twice. ¯\\\_(ツ)\_/¯

Anyway, as this issue is not caused by react-testing-library after all, it can be closed.

kentcdodds:
I literally just discovered this minutes ago and was making an example for you: https://codesandbox.io/s/0xw7401o50

[...]

PR #440, opened 2019-08-13
RFC - manually flush microtasks in afterEach

threepointone:
This PR replaces calling async act() in afterEach, with a version that still flushes microtasks, but doesn't wrap with act().

Explanation: If there are any hanging microtasks, that means a test ran without asserting on a valid state. Any async portion that causes updates should be asserted on, or atleast resolved within the scope of a test. This PR should still fire missing act warnings for a test, but won't let it leak to the next one.

### DIFF
--- a/src/flush-microtasks.js
+++ b/src/flush-microtasks.js
@@ -1,3 +1,7 @@
+/* istanbul ignore file */
+// the part of this file that we need tested is definitely being run
+// and the part that is not cannot easily have useful tests written
+// anyway. So we're just going to ignore coverage for this file
 /**
  * copied from React's enqueueTask.js
  */
--- a/src/flush-microtasks.js
+++ b/src/flush-microtasks.js
@@ -1,0 +1,43 @@
+/**
+ * copied from React's enqueueTask.js
+ */
+
+let didWarnAboutMessageChannel = false
+let enqueueTask
+try {
+  // read require off the module object to get around the bundlers.
+  // we don't want them to detect a require and bundle a Node polyfill.
+  const requireString = `require${Math.random()}`.slice(0, 7)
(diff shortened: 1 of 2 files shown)


bcarroll22:
Can you give an example of when this would prevent an issue that the current implementation doesn't? Not that I disagree with this, I just want to better understand what it's fixing.

kentcdodds:
This looks fine. A tiny bit complicated, but I'm cool with it, especially since it's coming right out of React. We're getting a failure thanks to code coverage. I'm not too concerned about this file though, so you can add `/* istanbul ignore file */` to the top of the file and we'll just exclude it from the coverage report.

Fun fact: I added `/* istanbul ignore file */` support: https://github.com/istanbuljs/istanbuljs/pull/108 😄

kentcdodds:
Oh, also, THANK YOU SUNIL!

kentcdodds:
I got that updated. I'm pretty 👍 about this change.

kentcdodds:
@threepointone, as this is titled "RFC" I just want to be certain that you think this should be added or if this PR was more of an idea/suggestion. I'm for it personally.

threepointone:
I missed the conversation here, sorry. I think we should do this, yes. I'll rebase the PR later today.

LaurensBosscher:
> I missed the conversation here, sorry. I think we should do this, yes. I'll rebase the PR later today.

Hey @threepointone, if you're busy I could do the rebase?

threepointone:
Yeah could you take this over? Thanks in advance!

LaurensBosscher:
> Yeah could you take this over? Thanks in advance!

Sorry that it took so long, I've created a new PR with the squashed commit.

kentcdodds:
Closing in favor of #156 514

PR #454, opened 2019-08-16
fix: spread console error args when proxying

eventualbuddha:
**What**: This is a bug fix to ensure proxied `console.error` calls are passed through correctly.

**Why**: Without them, `console.log('Value:', 1)` would get logged as `['Value:', 1]` instead of `Value: 1`.

**How**: I'm not sure how this applies.

**Checklist**:

- [ ] Tests N/A
- [X] Ready to be merged


kentcdodds:
@all-contributors please add @eventualbuddha for code

Issue #442, opened 2019-08-14
[bug] act-compat assume console.error is called with a string


SBoudrias:
- `react-testing-library` version: 
- `react` version:
- `node` version:
- `npm` (or `yarn`) version:

### Relevant code or config:

Code is situated here: https://github.com/testing-library/react-testing-library/blob/master/src/act-compat.js#L32

```js
args[0].indexOf(
              'Warning: Do not await the result of calling ReactTestUtils.act',
            ) === 0
```

This assumes the first arg passed to `console.error` is a string; but the type could be different. For example, it can be an object `console.error({ error })`

### Suggested solution:

We could either:

1. Stringify the args before calling `indexOf`
2. Type check the arg first

alexkrolick:
> Type check the arg first

This seems fine, since we can only check for the substring when the input is a string itself

kentcdodds:
Agreed. Would you like to [make a pull request](http://makeapullrequest.com/) for this @SBoudrias?

ely-alamillo:
@kentcdodds I can go ahead and stringify the args for this.

rvdkooy:
@ely-alamillo you still taking this? I can also do this one?

ely-alamillo:
Why does this case exist if it's not doing anything ? https://github.com/testing-library/react-testing-library/blob/master/src/act-compat.js#L38

@kentcdodds

kentcdodds:
Because otherwise this would run:

https://github.com/testing-library/react-testing-library/blob/8881803ea50f773cf909764299e68f35e22478d1/src/act-compat.js#L45

PR #469, opened 2019-08-29
add type checking to asyncAct args

ely-alamillo:
**What**:
This PR fixes issue: #442  by adding type checking for args used in asyncAct function.

**Why**:
Current implementation assumes console.erorr was always called with a string even though it could be called with other types.

**How**:
Wrapped execution of code in a type check if statement to ensure operations are done only on string types.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/old-act.js
+++ b/src/__tests__/old-act.js
@@ -98,4 +98,22 @@ test('async act recovers from sync errors', async () => {
   `)
 })
 
+test('async act handles values that are not strings', async () => {
+  try {
+    await asyncAct(() => {
+      throw new Error({error: 'test error'})
+    })
+  } catch (err) {
+    console.error('call console.error')
+  }
+  expect(console.error).toHaveBeenCalledTimes(1)
+  expect(console.error.mock.calls).toMatchInlineSnapshot(`
+    Array [
+      Array [
+        "call console.error",
+      ],
+    ]
+  `)
+})
+
 /* eslint no-console:0 */
--- a/src/__tests__/old-act.js
+++ b/src/__tests__/old-act.js
@@ -97,13 +97,14 @@ test('async act recovers from sync errors', async () => {
 
 test('async act handles values that are not strings', async () => {
   try {
-    await asyncAct(() => {
+    await asyncAct(async () => {
       console.error({message: 'some message'})
+      await null
     })
   } catch (err) {
     console.error('call console.error')
   }
-  expect(console.error).toHaveBeenCalledTimes(3)
+  expect(console.error).toHaveBeenCalledTimes(2)
   expect(console.error.mock.calls).toMatchInlineSnapshot(`
     Array [
       Array [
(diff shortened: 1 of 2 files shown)


kentcdodds:
Thanks. What happens if console.error is called with a non-string? I believe that currently this will result in the originalConsoleError not getting called.

ely-alamillo:
That is correct. I can move the else out of the if block to have it run whenever there is no string.

src/__tests__/old-act.js
```diff
@@ -98,4 +98,22 @@ test('async act recovers from sync errors', async () => {
   `)
 })
 
+test('async act handles values that are not strings', async () => {
+  try {
+    await asyncAct(() => {
+      throw new Error({error: 'test error'})
```
rvdkooy: I think you have to do a `console.error({ message: 'some message' }`  insteadof throwing it, to mimic the problem

ely-alamillo: Changed the test to not throw error.

rvdkooy: This is correct 👍 

but you had to change the assertion of another test, which is an indication of a wrong implementation (you are still not calling the original console.error)

check this implementation for inpiration :-):
https://github.com/rvdkooy/react-testing-library/commit/9020b83ca047c6dadaaa1e70407b6146caf5cff9

@kentcdodds can you have a look if we're on the right track with this?

src/__tests__/old-act.js
```diff
@@ -32,18 +32,15 @@ test('async act works even when the act is an old one', async () => {
     console.error('sigil')
   })
   expect(console.error.mock.calls).toMatchInlineSnapshot(`
-        Array [
-          Array [
-            "sigil",
```
kentcdodds: It looks like this log is no longer happening. Could you fix your changes so you don't have to change this snapshot?

src/act-compat.js
```diff
@@ -28,19 +28,21 @@ function asyncAct(cb) {
...
+            ) {
+              // v16.8.6
+              isAsyncActSupported = false
+            } else if (
+              args[0].indexOf(
+                'Warning: The callback passed to ReactTestUtils.act(...) function must not return anything',
+              ) === 0
+            ) {
+              // no-op
+            }
```
kentcdodds:
```suggestion
            } else {
              originalConsoleError.apply(console, args)
            }
```

I think that should do it.

src/__tests__/old-act.js
ely-alamillo: Changed implementation to revert back to this snapshot.

kentcdodds:
It looks like this is still unresolved... That snapshot should not change at all.

rvdkooy:
@ely-alamillo check this PR for the solution:
https://github.com/testing-library/react-testing-library/pull/476
(you can take the code if you want)

kentcdodds:
Thank you @rvdkooy. I've merged your PR.

Thank you @ely-alamillo for doing the initial work here. Even though we weren't able to merge your PR, I'm still happy for your efforts in contributing to the project!

kentcdodds:
@all-contributors please add @ely-alamillo for code and tests

PR #476, opened 2019-09-03
fix: console.error doesn't have to be a string in act.compat.js

rvdkooy:
What:
This PR fixes issue: #442 by adding type checking for args used in asyncAct function.

Why:
Current implementation assumes console.erorr was always called with a string even though it could be called with other types.

How:
Wrapped execution of code in a type check if statement to ensure operations are done only on string types.

Checklist:

- [ ]  Documentation added to the docs site N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/old-act.js
+++ b/src/__tests__/old-act.js
@@ -99,14 +99,11 @@ test('async act recovers from sync errors', async () => {
 })
 
 test('async act can handle any sort of console.error', async () => {
-  try {
-    await asyncAct(async () => {
-      console.error({error: 'some error'})
-      await null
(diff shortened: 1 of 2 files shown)


kentcdodds:
I'm happy to merge either of these once they pass the build and all the current tests :)

rvdkooy:
actually found out that the `noop` scenario was never tested 
the previous 100% coverage was not really 100%

Added a test for this, coverage should be 100% again

kentcdodds:
@all-contributors please add @rvdkooy for code and tests

PR #489, opened 2019-09-23
feat: add pure bundle build

### LINKED ISSUE #486: /pure build throws in browsers not supporting async-await
More of a problem with `kcd-scripts` but I'm filing it here since the use case is more obvious. Feel free to transfer the issue.

- `react-testing-library` version: 9.1.4
- `react` version: N/A
- `node` version: N/A
- `npm` (or `yarn`) version: N/A

### Relevant code or config:

```js
import '@testing-library/react/pure'
```

### What you did:

Used `@testing-library/react` in IE11

### What happened:

Threw with `unexpected identifier` (due to async-await)

### Reproduction:

https://github.com/mui-org/material-ui/pull/17165/commits/1e54d497d94b68d3ea594ac83fa3d2e62ab8026c 

failing build: https://circleci.com/gh/mui-org/material-ui/113431

### Problem description:

`kcd-scripts build` will use babel but ignore any `browserslists` assuming the target is `node: 8`. [Only rollup and webpack bundles will load browserslist](https://github.com/kentcdodds/kcd-scripts/blob/bd3c762c0202ac3564bb6a329e3e9e972795b5e7/src/config/babelrc.js#L40)

### Suggested solution:

Add a browserslist targetting ie 11, node 8 and the last 3 versions of the evergreen browsers as well as:

No special case for babel `targets`. It should always use `.browserslistrc` since that supports a `node` entry as well. I think it didn't use to support a node version since I occasionally encounter a repository using browserslist but throwing with a node entry.

kentcdodds:
**What**: Adds a "pure" build for bundling with rollup (and parallelizes all build scripts)

**Why**: Closes #486

**How**: Used the environment variables supported by the kcd-scripts rollup config: https://github.com/kentcdodds/kcd-scripts/blob/bd3c762c0202ac3564bb6a329e3e9e972795b5e7/src/config/rollup.config.js so we can make the pure.js file an entry and set the output filename suffix (so it doesn't overwrite the existing build).

**Checklist**:

      [docs site](https://github.com/alexkrolick/testing-library-docs) N/A. I don't think the UMD is something we really document or need to honestly.
- [ ] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,11 @@
     "node": ">=8"
   },
   "scripts": {
-    "build": "kcd-scripts build && kcd-scripts build --bundle --no-clean",
+    "prebuild": "rimraf dist",
+    "build": "npm-run-all --parallel build:main build:bundle:main build:bundle:pure",
+    "build:main": "kcd-scripts build --no-clean",
+    "build:bundle:main": "kcd-scripts build --bundle --no-clean",
+    "build:bundle:pure": "cross-env BUILD_FILENAME_SUFFIX=.pure BUILD_INPUT=src/pure.js kcd-scripts build --bundle --no-clean",
     "lint": "kcd-scripts lint",
     "test": "kcd-scripts test",
     "test:update": "npm test -- --updateSnapshot --coverage",
@@ -48,9 +52,12 @@
   "devDependencies": {
     "@reach/router": "^1.2.1",
     "@testing-library/jest-dom": "^4.1.0",
+    "cross-env": "^6.0.0",
     "kcd-scripts": "^1.7.0",
+    "npm-run-all": "^4.1.5",
     "react": "^16.9.0",
-    "react-dom": "^16.9.0"
+    "react-dom": "^16.9.0",
+    "rimraf": "^3.0.0"
   },
   "peerDependencies": {
     "react": "*",


eps1lon:
This doesn't work since `@testing-library/react/pure` uses 

```js
// ./pure.js
// makes it so people can import from '@testing-library/react/pure'
module.exports = require('./dist/pure')
```

May I ask why you don't use `.browserslistrc` for non-webpack and non-rollup builds? I guess this would require work for some of your packages that still use a `browserslist` version that don't support `node` entries?

kentcdodds:
I don't want to compile async/await for the majority of use cases (people running tests in a recent version of node with jsdom). I think that it's reasonable to expect people to do extra work for the edge case:

```javascript
import {render} from '@testing-library/react/dist/@testing-library/react.pure.esm.js'
```

If that's annoying enough, someone could make another module in their project that they import:

```javascript
// react-testing-library.js
export * from '@testing-library/react/dist/@testing-library/react.pure.esm.js'
export default from '@testing-library/react/dist/@testing-library/react.pure.esm.js'
```

and then

```javascript
import {render} from './react-testing-library'
```

eps1lon:
> version of node with jsdom). I think that it's reasonable to expect people to do extra work for the edge case:

Older browser support is not an edge case. It's directly in line with

> Interact with your app the same way as your users

My users don't use `jsdom`. They use a browser. And 2% of global users and more importantly 25% of screen reader users still use IE 11. I'll already do the extra work in making this runnable in a browser but waving this off as an "edge case" is a bit concerning. Especially if all the reason given for not compiling async/await is 

> I don't want to

> If that's annoying enough, someone could make another module in their project that they import:

I'll do that as well. If I was only concerned with my projects I wouldn't have opened this. I wanted to make this work as easy as possible for any `@testing-library/react` user.

I'm fine with having multiple bundles for various targets and happy with this change. Me personally I would make the bundle for older targets the default. Would like to hear some other opinions though.

kentcdodds:
I want to have the common-case be the default. Considering this library has been available for a very long time available at the current level of compilation and this is the first time we're hearing anyone want this tells me that the common case is definitely not to compile/polyfill things further.

eps1lon:
>Considering this library has been available for a very long time available at the current level of compilation 

async-await only got added recently. You can't consider the full time frame if the IE 11 target never would have had any effect.

PR #495, opened 2019-10-03
feat(debug): allow debugging an array of containers

danieljcafonso:
**What**: 
Adds the ability to log an array of containers.

**Why**:
Right now, if you try to debug the output of a queryAll or getAll, you will receive an error. This will allow everyone to see the output of each container inside of the output of that query.

**How**:
Checks if the received container is an array, if so, logs each element of it.
 
**Checklist**:
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/pure.js
+++ b/src/pure.js
@@ -60,8 +60,12 @@ function render(
   return {
     container,
     baseElement,
-    // eslint-disable-next-line no-console
-    debug: (el = baseElement) => console.log(prettyDOM(el)),
+    debug: (el = baseElement) =>
+      Array.isArray(el)
+        ? // eslint-disable-next-line no-console
+          el.forEach(e => console.log(prettyDOM(e)))
+        : // eslint-disable-next-line no-console,
+          console.log(prettyDOM(el)),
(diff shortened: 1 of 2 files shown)


src/pure.js
```diff
@@ -60,8 +60,12 @@ function render(
   return {
     container,
     baseElement,
-    // eslint-disable-next-line no-console
-    debug: (el = baseElement) => console.log(prettyDOM(el)),
+    debug: (el = baseElement) =>
+      Array.isArray(el)
+        ? // eslint-disable-next-line no-console
+          el.map(e => console.log(prettyDOM(e)))
```
kentcdodds:
```suggestion
          el.forEach(e => console.log(prettyDOM(e)))
```

We don't want to return an array with undefined values from `debug` :)

danieljcafonso: yeah, you're right :man_facepalming: Thank you Kent! I just applied those changes.

kentcdodds:
@all-contributors please add @danieljcafonso for code and tests

PR #505, opened 2019-10-23
chore(ci): add cron job install script for smoke testing next

kentcdodds:
**What**: Test react's `next` versions

**Why**: https://reactjs.org/blog/2019/10/22/react-release-channels.html#using-the-next-channel-for-integration-testing

**How**: added a step to the install in our `.travis.yml` which will upgrade react and react-dom to the `next` release if we're running the cron job.

**Checklist**:

- [ ] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,12 @@ node_js:
   - '8'
   - '10'
   - '12'
-install: npm install
+install:
+  - npm install
+  # as requested by the React team :)
+  # https://reactjs.org/blog/2019/10/22/react-release-channels.html#using-the-next-channel-for-integration-testing
+  - if [ "$TRAVIS_EVENT_TYPE" = "cron" ]; then npm install react@next
+    react-dom@next; fi
 script: npm run validate
 after_success: kcd-scripts travis-after-success
 branches:


kentcdodds:
I tested this locally and it works well. I've configured travis to have a cron job run every day. We should be set with these changes :)

kentcdodds:
Oh, and adding the cron job was really quick and painless. Used the docs here:

https://docs.travis-ci.com/user/cron-jobs

And here's what it looks like:

travis cron job showing it will run on master daily

.travis.yml
```diff
@@ -9,7 +9,12 @@ node_js:
   - '8'
   - '10'
   - '12'
-install: npm install
+install:
+  - npm install
+  # as requested by the React team :)
+  # https://reactjs.org/blog/2019/10/22/react-release-channels.html#using-the-next-channel-for-integration-testing
+  - if [ "$TRAVIS_EVENT_TYPE" = "cron" ]; then npm install react@next
+    react-dom@next; fi
```
kentcdodds: Whoops! Somehow this ended up on two lines 🙃 Maybe it'll work anyway, but I'll fix that. Should be on one line.

BTMPL: You can manually add a `\` at the end of line to signify multi-line statements.

Going to assume your prettier messed it up.

kentcdodds: Prettier doesn't typically introduce bugs. I just ran it locally and checked and prettier didn't touch this line. I'm pretty sure this happened because I just used vim and probably hit a key by mistake.

PR #507, opened 2019-10-24
feat: Add option to render concurrent roots

eps1lon:
First draft. There's a lot to figure out (like error message if the experimental build isn't used or if this cleanup is correct etc) 

**What**:

Adds `root` option to `render` which currently only allows `concurrent` but could be used to create blocking roots instead

**Why**:

hype-driven-development

**How**:

Added a lot of branching logic but overall the implementation would be simpler with `create*Root`.

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,8 +55,8 @@
     "cross-env": "^6.0.0",
     "kcd-scripts": "^1.7.0",
     "npm-run-all": "^4.1.5",
-    "react": "^16.9.0",
-    "react-dom": "^16.9.0",
+    "react": "^0.0.0-experimental-f6b8d31a7",
+    "react-dom": "^0.0.0-experimental-f6b8d31a7",
     "rimraf": "^3.0.0"
   },
   "peerDependencies": {
--- a/tests/setup-env.js
+++ b/tests/setup-env.js
@@ -1,1 +1,2 @@
 import '@testing-library/jest-dom/extend-expect'
+jest.mock('scheduler', () => require('scheduler/unstable_mock'))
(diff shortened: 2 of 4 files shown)


tests/setup-env.js
```diff
@@ -1 +1,2 @@
 import '@testing-library/jest-dom/extend-expect'
+jest.mock('scheduler', () => require('scheduler/unstable_mock'))
```
eps1lon: Biggest unknown for to figure out. I think we need to use this to advance time (or something like that) if we have tests scheduling work (like setState).

FredyC:
I wonder why are tests failing to find `createRoot` method. The package.json is correctly updated, there is no lock file to update, so it's rather weird.

package.json
```diff
@@ -55,8 +55,8 @@
     "cross-env": "^6.0.0",
     "kcd-scripts": "^1.7.0",
     "npm-run-all": "^4.1.5",
-    "react": "^16.9.0",
-    "react-dom": "^16.9.0",
+    "react": "^0.0.0-experimental-f6b8d31a7",
+    "react-dom": "^0.0.0-experimental-f6b8d31a7",
```
kentcdodds: I think that this resolves to the latest stable version. If we want to support the experimental release we have to do an exact version rather than `^`. I believe that's why tests are failing.

kentcdodds:
I'm not sure how close concurrent mode is (but I get the impression that it's not very close) so I think we should just close this for now and open it when concurrent mode is closer.

Thanks!

eps1lon:
Yeah that was me playing around. Not sure how testing will look with a mocked scheduler.

PR #514, opened 2019-10-30
manually flush microtasks in afterEach

LaurensBosscher:
This is the rebase of #440. I've squashed both commits into the original commit.

Original description:

> This PR replaces calling async act() in afterEach, with a version that still flushes microtasks, but doesn't wrap with act().
> 
> Explanation: If there are any hanging microtasks, that means a test ran without asserting on a valid state. Any async portion that causes updates should be asserted on, or atleast resolved within the scope of a test. This PR should still fire missing act warnings for a test, but won't let it leak to the next one.
>

### DIFF
--- a/src/flush-microtasks.js
+++ b/src/flush-microtasks.js
@@ -1,0 +1,47 @@
+/* istanbul ignore file */
+// the part of this file that we need tested is definitely being run
+// and the part that is not cannot easily have useful tests written
+// anyway. So we're just going to ignore coverage for this file
+/**
+ * copied from React's enqueueTask.js
+ */
+
+let didWarnAboutMessageChannel = false
+let enqueueTask
+try {
+  // read require off the module object to get around the bundlers.
+  // we don't want them to detect a require and bundle a Node polyfill.
+  const requireString = `require${Math.random()}`.slice(0, 7)
+  const nodeRequire = module && module[requireString]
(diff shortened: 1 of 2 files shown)


kentcdodds:
@all-contributors please add @LaurensBosscher code

evoyy:
I just upgraded to 9.3.1 and this is causing my test suite to log the following error:

ERROR: true, 'This browser does not have a MessageChannel implementation, so enqueuing tasks via await act(async () => ...) will fail. Please file an issue at https://github.com/facebook/react/issues if you encounter this warning.'

Is this something I need to worry about? I'm running my tests in Chrome using Karma.

kentcdodds:
cc @threepointone

donaldpipowitch:
This was the only change from v9.3.0 to v9.3.1, right? When I switch between those versions I get ` An update to Formik inside a test was not wrapped in act(...).` in every case where a `fireEvent.change` is not followed by an `await`. `fireEvent.click` seems to be fine. Is this intended behavior?

eps1lon:
@donaldpipowitch Please open a new issue with a reproducible example. Without code it's hard to tell whether this is intended or not.

donaldpipowitch:
@eps1lon I created the bug report: https://github.com/testing-library/react-testing-library/issues/535

PR #519, opened 2019-11-01
feat: Flush microtasks in cleanup

eps1lon:
BREAKING:

`cleanup` is now async.

```diff
- cleanup();
+ await cleanup();
```

**What**:

Flushes microtasks (introduced in #514) in cleanup instead of afterEach

**Why**:

Flushing microtask is not available for users of the `/pure` build.

**How**:

Move logic from index#afterEach into pure#cleanup

**Checklist**:

- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/cleanup.js
+++ b/src/__tests__/cleanup.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import {render, cleanup} from '../'
 
-test('cleans up the document', () => {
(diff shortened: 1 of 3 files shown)


kentcdodds:
We've got a few breaking changes to make in DOM Testing Library that we really need to push out. We should do all of this at the same time. Hopefully will get it done next week.

PR #540, opened 2019-12-04
fix: support automatic types in pnpm installs

aleclarson:
When installed with pnpm, the `@types/testing-library__react` package cannot be used by packages that don't directly depend on it, so we need to re-export them.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,1 @@
+export * from 'testing-library__react'


eps1lon:
I can't say anything to this particular change without an example repo.

It seems weird that pnpm requires this. It basically means that every package maintainer (which might have nothing to do with types/) need to add an index.d.ts now. 

This adds a lot of friction. Honestly I'd rather remove the types dependency with the next breaking change so that we handle them just like any other package: If you want types and the package does not bundle them then you have to install them.

bcarroll22:
Hmm yeah why is the types package a dependency? Would removing the dependency even be a breaking change? I don't think anything references them currently inside the package.

eps1lon:
> Hmm yeah why is the types package a dependency? Would removing the dependency even be a breaking change? I don't think anything references them currently inside this package.

It's used so that when installing `@testing-library/react` you don't need to install the types packages. It's automatically included.

But after thinking a bit more about this it's still not obvious why this change is necessary. I'm a bit worried this causes type conflicts. @aleclarson A complete repro would help a lot here.

aleclarson:
Bam https://github.com/aleclarson/repro/tree/rtl-540

aleclarson:
> It basically means that every package maintainer (which might have nothing to do with types/) need to add an index.d.ts now.

You might be confused. This pnpm incompatibility is caused by `@types/testing-library__react` being a transient dependency.

eps1lon:
> > It basically means that every package maintainer (which might have nothing to do with types/) need to add an index.d.ts now.
> 
> You might be confused. This pnpm incompatibility is caused by `@types/testing-library__react` being a transient dependency.

I guess you mean transitive.

Thanks for the repro. So basically simply adding the `@types/` package isn't enough for zero-config types in pnpm because transitive dependencies are not hoisted up which would be the case for npm or yarn v1.

Tested with module hoisting as well as ts 3.6 and it seemed to work. 

Friendly ping to @johnnyreilly since he had a [great blog post that explained the strategy](https://blog.johnnyreilly.com/2019/08/symbiotic-definitely-typed.html). I think the pnpm caveat should be mentioned.

Aside: I'll check out the kcd-scripts related failure.

eps1lon:
Since we don't have a lockfile CI stays red until https://github.com/kentcdodds/kcd-scripts/issues/107 is resolved.

kentcdodds:
I think I've fixed the issue in kcd-scripts. We we'll just need to upgrade that and we should be good to go.

eps1lon:
@kentcdodds The after_success script is always failing for fork PRs which means there will be no codecov report. However, this is a required check which means PRs can only be merged by admins not just members. Same issue for dom-testing-library.

kentcdodds:
Thanks for tuning me into that @eps1lon! I've been wondering for months why the codecov bot wasn't showing up on PRs. Now I've finally figured it out. This should fix that: https://github.com/kentcdodds/kcd-scripts/commit/ab38c66204e5e3c1e641e3558979a67b196be414

Is this ready to be merged?

PR #552, opened 2019-12-16
startsWith instead of indexOf

Connormiha:
**What**:
Just simplified checking for error name

**Why**:

**How**:
`indexOf` -> `startsWith`

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/act-compat.js
+++ b/src/act-compat.js
@@ -31,17 +31,13 @@ function asyncAct(cb) {
           const firstArgIsString = typeof args[0] === 'string'
           if (
             firstArgIsString &&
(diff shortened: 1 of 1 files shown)


Connormiha:
@eps1lon if IE11 supported, you right. This PR will break IE11.

eps1lon:
> @eps1lon if IE11 supported, you right. 

I don't think we officially support it as in "if it doesn't work in IE 11 we won't prioritize it i.e. PRs welcome". At least I'm using it in IE 11 and am therefore interested in it working as good as possible.

PR #553, opened 2019-12-17
docs: Fixed example code to use "screen"

sakito21:
Fixed example code to use `screen`.


kentcdodds:
@all-contributors please add @sakito21 for docs

wooller:
Why was this added? Im struggling to find reference to screen in react-testing-library outside of this doc change. Typescript also complains that this is not exported :/

PR #560, opened 2019-12-29
Fix error on complex example

tteke:
**What**: Fixing an error on readme

**Why**: In the complex example catch already sets error to ```error.message```. So in render, it should just show ```state.error```.

**How**: Render function updated to just show ```state.error``` instead of ```state.error.message```.

**Checklist**:

- [ ] Tests N/A
- [x] Ready to be merged
      
Yes. Either that or catch shoud update error on the state as ```error: error``` instead of ```error: error.message```

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,7 +248,7 @@ function Login() {
         </div>
         <button type="submit">Submit{state.loading ? '...' : null}</button>
       </form>
-      {state.error ? <div role="alert">{state.error.message}</div> : null}
+      {state.error ? <div role="alert">{state.error}</div> : null}
       {state.resolved ? (
         <div role="alert">Congrats! You're signed in!</div>
       ) : null}


README.md
```diff
@@ -248,7 +248,7 @@ function Login() {
         </div>
         <button type="submit">Submit{state.loading ? '...' : null}</button>
       </form>
-      {state.error ? <div role="alert">{state.error.message}</div> : null}
+      {state.error ? <div role="alert">{state.error}</div> : null}
```
dkarski: Hi, I have a question about the above syntax.

Is it some convenience to handle show/hidden a component like in README.md?
`{ isTrue ? <Component /> : null }` than `{ isTrue && <Component /> }` (shorter a way)

kentcdodds: I could write a blog post about this for the number of times I talk about it 😅

Basically I've been bitten by this before:

```
{contacts.length && <RenderContactList contacts={contacts} />}
```

If `contacts.length` is `0` then you'll render `0`. No good. Shipped that to prod at PayPal. So I never do that and instead I always go with the more explicit version:

```
{contacts.length ? <RenderContactList contacts={contacts} /> : null}
```

But you do whatever you like :)

kentcdodds:
@all-contributors please add @tteke for docs

tteke:
Thanks

Issue #575, opened 2020-01-21
Drop Node 8


kentcdodds:
Because [it's no longer LTS](https://nodejs.org/en/about/releases/).

thomasgrz:
This would be resolved by updating [to >= Node 10](https://github.com/testing-library/react-testing-library/blob/d6fac0584cd06d39123301c8c2ed33ca5425a2f3/package.json#L8), correct? Happy to submit a PR for this as this is a project I use fairly frequently.

eps1lon:
> I assume there may be some security settings on the repo to disallowed it?

You usually set the `origin` to the repo you own and `upstream` to the fork. See https://gist.github.com/Chaser324/ce0505fbed06b947d962 or https://help.github.com/en/github/getting-started-with-github/fork-a-repo.

thomasgrz:
@eps1lon thank you! I believe I properly (?) submitted the PR thanks to your guidance 👍 .

PR #576, opened 2020-01-23
Drop Node 8 support

thomasgrz:
**What**:
Updating the engine to Node v.10.

**Why**:
Node v.8 no longer has LTS. Closes https://github.com/testing-library/react-testing-library/issues/575

**How**:
Updating the engine field in `package.json` & `.travis.yml` respectively.

**Checklist**:

- [N/A] Documentation added to the
- [N/A] Tests
- [N/A] Typescript definitions updated
- [X] Ready to be merged

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,9 @@ cache:
 notifications:
   email: false
 node_js:
-  - '8'
   - '10'
   - '12'
+  - '13'
 install:
   - npm install
   # as requested by the React team :)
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,9 +6,9 @@ cache:
 notifications:
   email: false
 node_js:
-  - '8'
   - '10'
   - '12'
+  - '13'
```
hansjhoffman: The Node.js folks don't recommend v13 for production since it wont be promoted to LTS. Maybe both just v10 and v12 will suffice?

eps1lon: It's the testing environment not the production environment. If this adds too much work we can remove it. Otherwise it seems useful to monitor support.

Issue #577, opened 2020-01-24
fireEvent - mouseEnter/mouseLeave not working with addEventListener


greypants:
- `react-testing-library` version: 9.4.0
- `react` version: 16.12.0
- `node` version: 10.16.0
- `npm` (or `yarn`) version: 1.21.1

### Relevant code or config:

```jsx
import React, { useEffect, useState, useRef } from "react";
import { render, fireEvent } from "@testing-library/react";

const HoverMe = () => {
  const ref = useRef();
  const [isMouseEntered, setIsMouseEntered] = useState(false);

  useEffect(() => {
    const setYes = () => setIsMouseEntered(true);
    const button = ref.current;
    // If you change the event to "mouseover"
    // the test passes, even if you don't update the test
    button.addEventListener("mouseenter", setYes);

    return () => {
      button.removeEventListener("mouseenter", setYes);
    };
  });

  return <button ref={ref}>{isMouseEntered ? "yes" : "no"}</button>;
};

describe("mouseenter/mouseleave bug", () => {
  test("mouseenter should update text to 'yes'", () => {
    const { getByText } = render(<HoverMe />);
    fireEvent.mouseEnter(getByText("no"));
    expect(getByText("yes")).toBeTruthy();
  });
});
```

### What you did:

I'm using a third party library that attaches `mouseenter` and `mouseleave` events to a DOM element accessed via a React `ref` via `HTMLElement.addEventListener`. Was testing this behavior.

### What happened:

Works fine in the browser, but when writing my tests, I noticed that `fireEvent.mouseEnter` and `fireEvent.mouseLeave` have no effect. What's weird is if I leave the test alone, but change the event in the component to `mouseover/mouseout`, it works. If I attach the events the react way, via `onMouseEnter` and `onMouseLeave`, it also works.

### Reproduction:

https://codesandbox.io/s/react-testing-library-demo-37yqv?fontsize=14&hidenavigation=1&theme=dark

### Problem description:

`fireEvent.mouseLeave` and `fireEvent.mouseOver` do not work when the events are added via `addEventListener`.

### Suggested solution:

It may have something to do with this: https://reactjs.org/docs/events.html#mouse-events

If this is a React limitation, or if there's a workaround, it would be great to add it to the FAQ.

greypants:
[...]

PR #588, opened 2020-02-12
fix: make fireEvent mouseEnter/mouseLeave work with addEventListener

zbrogz:
(#577)

**What**:
Make `fireEvent.mouseEnter` and `fireEvent.mouseLeave` work with events added using `addEventListener`.

**Why**:
There is no way to trigger a native mouseEnter/mouseLeave event using React Testing Library. A workaround is to manually use `dispatchEvent`, but it's not as convenient as `fireEvent`. The reason for this limitation is due to how React handles mouseEnter/mouseLeave: https://github.com/facebook/react/issues/12978

**How**:
Instead of setting `fireEvent.mouseEnter/Leave` to `fireEvent.mouseOver/Out`, set `fireEvent.mouseEnter/Leave` to call both `mouseEnter/Leave` and `fireEvent.mouseOver/Out`. 

This makes the behavior match other events that can be called with `fireEvent`. For example, if one click handler was added to an element with `addEventListener` and another was added to the component directly with `<button onClick={handleClick}>...`, `fireEvent.click(button)` calls both click handlers.

**Checklist**:

- [ ] Documentation added to the N/A
- [ ] Tests N/A
- [x] Ready to be merged
      

Credit to @jessethomson

### DIFF
--- a/src/__tests__/events.js
+++ b/src/__tests__/events.js
@@ -128,9 +128,6 @@ const eventTypes = [
   },
 ]
 
-// native select event isn't passing
-const nonNativeEvents = {doubleClick: 'dblclick', select: true}
-
 eventTypes.forEach(({type, events, elementType, init}) => {
   describe(`${type} Events`, () => {
     events.forEach(eventName => {
@@ -160,11 +157,10 @@ eventTypes.forEach(({type, events, elementType, init}) => {
   describe(`Native ${type} Events`, () => {
     events.forEach(eventName => {
       let nativeEventName = eventName.toLowerCase()
-      if (nonNativeEvents[eventName]) {
-        if (nonNativeEvents[eventName] === true) {
-          return
-        }
-        nativeEventName = nonNativeEvents[eventName]
+
+      // The doubleClick synthetic event maps to the dblclick native event
+      if (nativeEventName === 'doubleclick') {
+        nativeEventName = 'dblclick'
       }
 
       it(`triggers native ${nativeEventName}`, () => {
--- a/src/__tests__/events.js
+++ b/src/__tests__/events.js
@@ -128,6 +128,9 @@ const eventTypes = [
   },
 ]
 
+// native select event isn't passing
+const nonNativeEvents = {doubleClick: 'dblclick', select: true}
+
 eventTypes.forEach(({type, events, elementType, init}) => {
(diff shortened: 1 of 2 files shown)


weyert:
Cool, thanks, @zbrogz, well done :D

zbrogz:
I'll update this with tests.

kentcdodds:
Thank you so much for this. I would love a test as well.

Is there a real world scenario where people would have to callbacks called unexpectedly?

zbrogz:
I added tests to verify that all native events can be triggered using `fireEvent`.  A couple caveats: 

1. Native `onSelect` isn't passing for some reason. 
2. `onDoubleClick` maps to `dblclick`.

If the scope of this is too big, I'll remove this and just write a test limited to onMouseEnter/onMouseLeave, like mentioned earlier.

zbrogz:
> Is there a real world scenario where people would have two callbacks called unexpectedly?

I don't believe so. This makes it so `fireEvent.onMouseLeave/Enter` is consistent with the way other events work: both the native (added with `addEventListener`) and react (passed using 'onEventName`  prop) event handlers should get called.

zbrogz:
@kentcdodds Do you mind taking a look at this again?

src/__tests__/events.js
```diff
@@ -128,6 +128,9 @@ const eventTypes = [
   },
 ]
 
+// native select event isn't passing
+const nonNativeEvents = {doubleClick: 'dblclick', select: true}
```
kentcdodds: Should `doubleclick` even be in the events list?

Also, `select` probably doesn't work because JSDOM, which is unfortunate.

I probably wouldn't implement things this way and instead I would just have if statements in the test itself for each of these individually with comments explaining why these are special cases. No reason to have this indirection/abstraction.

zbrogz: @kentcdodds I made some changes based on your feedback.
>instead I would just have if statements in the test itself for each of these individually with comments explaining why these are special cases
Done
> Should doubleclick even be in the events list?
I think it makes sense to test both the doubleClick (synthetic) and dblclick (native) events. 
> Also, select probably doesn't work because JSDOM, which is unfortunate.
I included a fix for this in the latest commit.

kentcdodds:
@all-contributors please add @zbrogz for code and tests

PR #598, opened 2020-03-04
Remove cleanup-after-each

MichaelDeBoey:
It was deprecated in the last release, so we could safely delete it now I think 🙂


kentcdodds:
@all-contributors please add @MichaelDeBoey for code

PR #619, opened 2020-03-21
Update dependencies + use kcd-scripts' husky

MichaelDeBoey:


package.json
```diff
@@ -41,21 +36,21 @@
     "end-to-end",
     "e2e"
   ],
-  "author": "Kent C. Dodds <me@kentcdodds.com> (https://kentcdodds.com)",
+  "author": "Kent C. Dodds <kentcdodds@gmail.com> (https://kentcdodds.com)",
```
kentcdodds: I'd prefer this to be: me@kentcdodds.com

MichaelDeBoey: I just copied it over from `generator-kcd-oss`, will change it again 🙂

PR #628, opened 2020-04-03
docs: Fix simple typo, seemless -> seamless

timgates42:
There is a small typo in src/act-compat.js.

Should read `seamless` rather than `seemless`.


afontcu:
Thanks! :)

PR #645, opened 2020-04-21
docs(readme): Add install command for yarn

Ishaan28malik:
**What**: Added about the yarn installation process.

**Why**: For developers who run the build using yarn 

**How**: Adding the yarn add bash command.

**Checklist**:

- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ or
 
 for installation via [yarn](https://classic.yarnpkg.com/en/)
 ```
-yarn add --dev jest @testing-library/react @testing-library/user-event jest-dom nock
+yarn add --dev @testing-library/react
 ```
 
 This library has `peerDependencies` listings for `react` and `react-dom`.
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ npm install --save-dev @testing-library/react
 ```
 or
 
-This module is distributed via [yarn](https://classic.yarnpkg.com/en/)
+for installation via [yarn](https://classic.yarnpkg.com/en/)
 ```
 yarn add --dev jest @testing-library/react @testing-library/user-event jest-dom nock
 ```
--- a/README.md
+++ b/README.md
@@ -101,6 +101,12 @@ should be installed as one of your project's `devDependencies`:
 ```
 npm install --save-dev @testing-library/react
 ```
+or
+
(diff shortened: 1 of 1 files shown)


Ishaan28malik:
@afontcu Please review this.

README.md
```diff
@@ -101,6 +101,12 @@ should be installed as one of your project's `devDependencies`:
 ```
 npm install --save-dev @testing-library/react
 ```
+or
+
+This module is distributed via [yarn](https://classic.yarnpkg.com/en/)
```
eps1lon, line 106: yarn isn't a distribution channel but a package manager. What do you want to express with that statement?

README.md
```diff
@@ -101,6 +101,12 @@ should be installed as one of your project's `devDependencies`:
...
+```
+yarn add --dev jest @testing-library/react @testing-library/user-event jest-dom nock
```
eps1lon, line 108: I would move this to the npm block above so that it's clearer that these commands do the same thing. It should also install the same dependencies. `user-event`, `jest-dom` and `nock` are not required to use this library.

README.md
Ishaan28malik: Oh yeah, you are right, I just made a mistake there.

README.md
Ishaan28malik: Sure, my only idea behind this PR was to add the yarn installation method as well to make the user stick to this particular repo and docs rather than making random search over web.

eps1lon: It's still unclear why you think `@testing-library/user-event jest-dom nock` are required for usage with `yarn`. I use none of these libraries and `yarn` and hadn't had any issue with missing dependencies.

Ishaan28malik: Oh okay so, what would be added for users who are using ```yarn``` and need a solution for it , because yarn add --save-dev @testing-library/react seems okay to me, let me update the changes. 

Sorry, but I followed an online article that uses the previous command that updated for yarn installation.

Ishaan28malik:
@eps1lon do I still need to make any changes?

eps1lon:
@all-contributors add @Ishaan28malik  for docs

eps1lon:
@Ishaan28malik Thanks!

Ishaan28malik:
@kentcdodds that's great to know.

PR #656, opened 2020-05-04
Return value from debug for test runners like Wallaby.js

deshiknaves:
**What**:

When using [Wallaby.js](https://wallabyjs.com) there is currently no way to get the output of the `debug` function in the interface. This is because the `debug` function just logs the changes and doesn't return anything. This PR will return a `string` or an `array` of `string`s from the `debug` function, which will enable:

image

**Why**:

The `debug` function has been updated to `console.log` the output first — maintaining the current functionality. And then return a `string` or an `array` of `string`s using `prettyDom` without the `highlight` option.

**How**:
**Checklist**:

      [docs site](https://github.com/alexkrolick/testing-library-docs) N/A (I don't think this needs an update)
- [x] Tests
- [ ] Ready to be merged
      

I have pulled down the `DefinitelyTyped` repo and made the changes there — [pr/branch](https://github.com/deshiknaves/DefinitelyTyped/tree/pr/testing-library-react-debug-return-type). I'm a little bit unsure of which order the PRs should go in. I'm thinking this should be merged first and then the update to `DefinitelyTyped`? Let me know if I'm wrong and I can do whatever is required.

### DIFF
--- a/src/__tests__/debug.js
+++ b/src/__tests__/debug.js
@@ -23,13 +23,13 @@ test('debug returns the container', () => {
   const HelloWorld = () => <h1>Hello World</h1>
   const {debug} = render(<HelloWorld />)
   expect(debug()).toMatchInlineSnapshot(`
-    "[36m<body>[39m
-      [36m<div>[39m
-        [36m<h1>[39m
-          [0mHello World[0m
-        [36m</h1>[39m
-      [36m</div>[39m
-    [36m</body>[39m"
+    "<body>
+      <div>
+        <h1>
+          Hello World
+        </h1>
+      </div>
+    </body>"
   `)
 })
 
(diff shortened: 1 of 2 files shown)


kentcdodds:
Oh yeah, we should just put the prettyDOM call into a variable so we don't have to call it twice

deshiknaves:
@kentcdodds that's what I had to begin with, the problem is the highlighting:

image

The rationale was that `debug` is used when writing the test and not often when that is done, so I figured a little extra time would not be the worst.

However, @eps1lon I completely agree 99%. No, wallaby has no outputs for `console.log`s just doesn't show up anywhere. I reached out the the developer and they rely on a return value.

If I were to create a separate function, what would be an appropriate name?

kentcdodds:
I think it'd be better if instead of using `debug` for this, you simply make your own function that calls `prettyDOM(document.body, maxLength, { ...options, highlight: false })` and call that. I don't think I want to make the API to Testing Library any more complicated to support this edge case. Sorry :-/

PR #676, opened 2020-05-27
Run cleanup() when test runner supports teardown()

screendriver:
**What**:

`cleanup()` now runs not only when `afterEach()` exists. It also runs when `teardown()` exists.

**Why**:

There are test runners that either don't have a `afterEach()` function or supports different UI's, like [Mocha's TDD UI](https://mochajs.org/#tdd) for example. It would be great if `react-testing-library` would also support `teardown()` for cleaning up things.

**How**:

I just added an additional check that checks and calls `teardown()` instead of `beforeEach()`.

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged
      

The main problem that I have with this pull request is:

- linting fails because `teardown` is not defined. Is this somewhere deep in `eslint-config-kentcdodds`?
- I had problems to write a test for that because you are using `Jest` which does not have `teardown()`

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,6 +6,8 @@ import {cleanup} from './pure'
 // if you don't like this then either import the `pure` module
 // or set the RTL_SKIP_AUTO_CLEANUP env variable to 'true'.
 if (!process.env.RTL_SKIP_AUTO_CLEANUP) {
+  // ignore teardown() in code coverage because Jest does not support it
+  /* istanbul ignore else */
   if (typeof afterEach === 'function') {
     afterEach(async () => {
       await cleanup()
--- a/src/index.js
+++ b/src/index.js
@@ -11,6 +11,9 @@ if (!process.env.RTL_SKIP_AUTO_CLEANUP) {
       await cleanup()
     })
   } else if (typeof teardown === 'function') {
+    // Block is guarded by `typeof` check.
+    // eslint does not support `typeof` guards.
+    // eslint-disable-next-line no-undef
     teardown(async () => {
       await cleanup()
     })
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -1,14 +1,20 @@
...
-  afterEach(async () => {
-    await cleanup()
-  })
+if (!process.env.RTL_SKIP_AUTO_CLEANUP) {
+  if (typeof afterEach === 'function') {
+    afterEach(async () => {
+      await cleanup()
+    })
+  } else if (typeof teardown === 'function') {
+    teardown(async () => {
```
eps1lon, line 14: I think you can just disable `no-undef` for this line and add an explainer that eslint doesn't support `typeof` checks:
```suggestion
    // Block is guarded by `typeof` check.
    // eslint does not support `typeof` guards.
    // eslint-disable-next-line no-undef
    teardown(async () => {
```

screendriver:
Now the tests fail because code coverage is below the threshold. Wich leads to my second point

- I had problems to write a test for that because you are using Jest which does not have teardown()

eps1lon:
I think you can tell coverage to ignore certain branches. Something like `instanbul ignore-if` or so. Should be able to find this via google "jest coverage ignore".

screendriver:
@eps1lon I ignored the `else` block and the build is now green

screendriver:
> I wonder whether we should consider this a breaking change 😬

Good question. I guess not? I mean, nothing will break. It's "just" a new feature for (for example) `Mocha` users like me 🙃

kentcdodds:
@all-contributors please add @screendriver for code

PR #696, opened 2020-06-05
fix(TS): add type defs file for pure

### LINKED ISSUE #695: No type definitions for react/pure
- `@testing-library/react` version: 10.2.0
- Testing Framework and version: jest@25.4.0
  
- DOM Environment: jsdom@15
  

### Relevant code or config:

```ts
import { render } from "@testing-library/react/pure";
```

### What happened:

> TS7016: Could not find a declaration file for module '@testing-library/react/pure'. 

### Reproduction:

https://codesandbox.io/s/nifty-tharp-3de0h?file=/src/index.ts:61-114

### Problem description:

This is happening because the `pure.d.ts` file is not next to `pure.js`, it's in a `types` folder and TS doesn't know how to link it. It's OK for the entry index because there's a `pkg.types` pointing to it, but importing something else requires the type file to be colocated.

### Suggested solution:

I think the files could either be moved next to their corresponding `.js` files and the build step updated to include them, or have a separate build step that does `cp types/*.d.ts dist/`.

kentcdodds:
Closes #695

### DIFF
--- a/pure.d.ts
+++ b/pure.d.ts
@@ -1,0 +1,1 @@
+export * from './types/pure.d.ts'
--- a/pure.d.ts
+++ b/pure.d.ts
@@ -1,1 +1,1 @@
-export * from './types/pure.d.ts'
+export * from './types/pure'
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "dist",
     "dont-cleanup-after-each.js",
     "pure.js",
+    "pure.d.ts",
     "types"
   ],
   "keywords": [


pure.d.ts
```diff
@@ -0,0 +1 @@
+export * from './types/pure.d.ts'
```
NiGhTTraX, line 1: > TS2691: An import path cannot end with a '.d.ts' extension. Consider importing './types/pure' instead.

```suggestion
export * from './types/pure'
```

kentcdodds:
@all-contributors please add @NiGhTTraX for bugs and review

PR #704, opened 2020-06-08
fixed complex example so that it actually works

weisssean:
**What**: Fixed the example in the README.md file so it passes the tests (only touched that file)

**Why**: When I tried to use this library I implemented the README tests, needless to say, complex example has serious errors in it and doesn't work.

**How**: just used waitFor instead of async, defined a variable that was missing and used axios instead of fetch (your 500 error in fetch wasn't catching).

**Checklist**:

- [ x] Documentation added to the
- [ N/A] Tests
- [ N/A] Typescript definitions updated
- [ x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,13 +206,14 @@ test('shows the children when the checkbox is checked', () => {
 ```jsx
 // login.js
 import React from 'react'
+import axios from 'axios'
 
 function Login() {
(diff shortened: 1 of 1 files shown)


kentcdodds:
Thanks for your work on this! @kettanaito is actually working on improving these examples everywhere and they'll be a bit different, so I think I'll close this. Thanks anyway!

PR #705, opened 2020-06-08
Fix pretty format import

### LINKED ISSUE #694: Cannot find module 'pretty-format'
- `@testing-library/react` version: 10.2.0
- Testing Framework and version:
  * `@testing-library/dom` version 7.9.0
  * `jest` version 25.5.4
- DOM Environment:
  * `jsdom` version 15.2.1
  * `react` version 16.13.1
  * `node` version 13.6.0
  * `typescript` version 3.9.3

### What you did:

Dependabot upgrade of `@testing-library/react` from 10.0.4 to 10.2.0.

### What happened:

In CI, our npm script that invokes `tsc --noEmit` fails with the following:

```
➜  assets git:(dependabot/npm_and_yarn/assets/testing-library/react-10.2.0) npx tsc --noEmit
node_modules/@testing-library/react/types/index.d.ts:3:54 - error TS2307: Cannot find module 'pretty-format' or its corresponding type declarations.

3 import {OptionsReceived as PrettyFormatOptions} from 'pretty-format'
                                                       ~~~~~~~~~~~~~~~
```

### Problem description:

It can't find `pretty-format`. I do see `pretty-format` in my `package-lock.json`, version 25.5.0. It might be the "corresponding type declarations" that the error is flagging? I notice this line was added in #690, which finagles the type definitions a bit, though I don't totally understand the PR.

It could be that I'm supposed to update my package-json with a new dependency of some sort?

samtsai:
**What**: add `pretty-format` as a dependency

**Why**: fix #694, since moving typings to this repo we have to make `pretty-format` an explicit dependency since our types rely on it

**How**: update `package.json`

**Checklist**:

- [x] Tests N/A
- [x] Ready to be merged
      

I tried with `devDependency` but those packages will not be installed on consumer projects, https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies is a good discussion that helped me wrap my head around what is necessary.


samtsai:
> Shouldn't it come with `@testing-library/dom`? I'm worried about keeping the versions in sync.

This was my worry too, and maybe this issue is not for all users. I'll keep investigating with the OP repo.

samtsai:
I'm going to close this PR, I have given feedback to the OP.

nickmccurdy:
Seems like we do need this based on https://github.com/testing-library/react-testing-library/issues/694#issuecomment-647736937

PR #714, opened 2020-06-19
Added the ability to pass props to the wrapper component in render()

ryall:
**What**:

I have added an additional option to `render` called `wrapperProps` to allow props to be passed to the wrapper component on creation. By default this is `null` (as before), meaning no breaking changes.

**Why**:

I found the need to pass additional props to my custom provider wrapper as outlined in the docs. Specifically I am using the Apollo Client provider `MockedProvider`, which allows mock responses to be applied using the provider.

**How**:

Simply replaced `React.createElement(..., PROPS, ...)` with an option variable passed into render, rather than always being `null`.

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

Will update the docs if this request is approved.

### DIFF
--- a/src/__tests__/render.js
+++ b/src/__tests__/render.js
@@ -87,3 +87,30 @@ test('renders options.wrapper around node', () => {
 </div>
 `)
 })
+
+test('passes options.wrapperProps to wrapper', () => {
+  const WrapperComponent = ({divClass, children}) => (
+    <div className={divClass} data-testid="wrapper-with-prop">
+      {children}
+    </div>
+  )
+
+  const {container} = render(<div data-testid="inner" />, {
+    wrapper: WrapperComponent,
+    wrapperProps: {
+      divClass: 'passed-by-prop',
+    },
+  })
+
+  expect(screen.getByTestId('wrapper-with-prop')).toBeInTheDocument()
+  expect(container.firstChild).toMatchInlineSnapshot(`
+<div
+  class="passed-by-prop"
+  data-testid="wrapper-with-prop"
+>
+  <div
+    data-testid="inner"
+  />
+</div>
+`)
+})
--- a/src/pure.js
+++ b/src/pure.js
@@ -36,6 +36,7 @@ function render(
     queries,
     hydrate = false,
     wrapper: WrapperComponent,
+    wrapperProps = null,
   } = {},
 ) {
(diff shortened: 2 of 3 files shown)


nickmccurdy:
Since we're abstracting props anyway, what would you think about something for context and other parent components? For example, you often need context Providers like Redux's store Provider or React Router's Routers to wrap elements in test. We're currently recommending people write their own wrappers, but a first party API to simplify this global setup could be useful.

kentcdodds:
I'm not sure I'm in favor of this. You can accomplish what you want without changing the API at all with this:

```javascript
  const WrapperComponent = ({divClass, children}) => (
    <div className={divClass} data-testid="wrapper-with-prop">
      {children}
    </div>
  )

  const {container} = render(<div data-testid="inner" />, {
    wrapper: props => <WrapperComponent {...props} divClass="passed-by-prop" />,
  })
```

I prefer this because it's less API to worry about/document/explain/maintain.

ryall:
> I'm not sure I'm in favor of this. You can accomplish what you want without changing the API at all with this:
> 
> ```js
>   const WrapperComponent = ({divClass, children}) => (
>     <div className={divClass} data-testid="wrapper-with-prop">
>       {children}
>     </div>
>   )
> 
>   const {container} = render(<div data-testid="inner" />, {
>     wrapper: props => <WrapperComponent {...props} divClass="passed-by-prop" />,
>   })
> ```
> 
> I prefer this because it's less API to worry about/document/explain/maintain.

Sure that's a possible solution, but `wrapperProps` is only a minor addition and keeps the code a bit cleaner in my opinion, as you don't have to remember to apply the parent props in every instance.

Plus it doesn't cause issues when using `react/jsx-props-no-spreading`.

ryall:
In the end this is the solution I have gone with:

```ts
export const getRenderWithProviderWrapper = <P extends object, R extends object>(
  provider: React.ComponentType<P>,
  providerProps?: P | null,
  returnExtension?: R,
) => {
  return (element: React.ReactElement, options?: RenderOptions) => {
    const wrapper = (props) => React.createElement(provider, { ...props, ...providerProps });

    return {
      ...render(element, { wrapper, ...options }),
      ...returnExtension,
    } as RenderResult & R;
  };
};
```

PR #717, opened 2020-06-20
Link to new Discord

nickmccurdy:


README.md
```diff
@@ -429,7 +429,7 @@ For questions related to using the library, please visit a support community
 instead of filing an issue on GitHub.
 
 - [Spectrum][spectrum]
-- [Reactiflux on Discord][reactiflux]
+- [Discord][discord]
```
kentcdodds, line 432: Let's ditch the spectrum link too

nickmccurdy: Done, what about the badge though?

README.md
```diff
@@ -428,8 +428,7 @@ a 👍. This helps maintainers prioritize what to work on.
 For questions related to using the library, please visit a support community
 instead of filing an issue on GitHub.
```
afontcu, line 429: (Unrelated, but I think we're missing an "l" here)

```suggestion
instead of filling an issue on GitHub.
```

nickmccurdy: I think the usage here is intentionally "filing" as in "to file paperwork". "filling" would technically also work in this context but I feel like it has different semantics, it's more about completing an issue than submitting it.

afontcu: oh, right, that makes sense. thanks!

README.md
kentcdodds: Let's get rid of all spectrum badges. Is there a discord badge we can use? I've seen some that show how many are online

nickmccurdy: What do you prefer?

[Discord](https://discord.gg/c6JN9fM)
[Discord](https://discord.gg/c6JN9fM)

You can also customize further by clicking on the Discord badge here https://shields.io/category/chat

kentcdodds: Let's do the second one. Looks cooler.

kentcdodds:
Thank you! Let's do this in all the repos! 👍👍

nickmccurdy:
https://github.com/orgs/testing-library/projects/1#card-40465069

afontcu:
on it! :) https://github.com/testing-library/vue-testing-library/pull/148

PR #718, opened 2020-06-22
Add iFood to users

nickmccurdy:
Responding to this request we got in Reactiflux:

> Hi, there! I work at iFood, the largest foodtech delivery in Latin America. I want to understand how do you select the companies that appear in the section Who's Using This in the Testing Library because we would like to join that - our frontend team is a heavy user of your platform! Tks a lot, Larissa

__Note:__ The site appears to only be available in Brazilian Portuguese

### DIFF
--- a/other/USERS.md
+++ b/other/USERS.md
@@ -3,10 +3,10 @@
 If you or your company uses this project, add your name to this list! Eventually
 we may have a website to showcase these (wanna build it!?)
 
-> No users have been added yet!
-
 <!--
 This file should just be a bulleted list like this:
 
 - [Company/Project/Person](https://example.com) uses it in [some app](https://example.com)
 -->
+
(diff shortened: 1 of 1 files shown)


kentcdodds:
I think they wanted to appear on the website.

nickmccurdy:
I got confused because there only seem to be users pages in the repositories, not the site. Should the site have one for all libraries?

kentcdodds:
I'm pretty sure they want to be on this list (on the home page of https://testing-library.com)

image

https://github.com/testing-library/testing-library-docs/blob/6ca45677fcd7c91c5e571b760906a445aaa287ec/website/siteConfig.js

nickmccurdy:
Oh, I didn't notice the wording there, my bad 😓

PR #720, opened 2020-06-24
fix(cleanup): microtask flushing now supports fake timers

kentcdodds:
**What**: microtask flushing now supports fake timers

**Why**: So people can use fake timers in their react tests without it causing issues with React's scheduler

**How**:

- Three days of total pain and anguish...
- Detect whether we're using fake timers and if we are, advance them forward.

**Checklist**:

~- [ ] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs)~ N/A
~- [ ] Tests~ N/A (really really hard to reproduce)
~- [ ] Typescript definitions updated~ N/A
- [x] Ready to be merged

### DIFF
--- a/src/flush-microtasks.js
+++ b/src/flush-microtasks.js
@@ -3,8 +3,17 @@
 // and the part that is not cannot easily have useful tests written
 // anyway. So we're just going to ignore coverage for this file
 /**
- * copied from React's enqueueTask.js
+ * copied and modified from React's enqueueTask.js
  */
+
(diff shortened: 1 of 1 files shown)


9still:
@kentcdodds Just stumbled on this - Thanks for fixing this!  FYI: I just filed https://github.com/facebook/jest/issues/10555 to hopefully get jest to provide an official way to find out if fake timers are being used.  Just referencing this here so if the API *is* added there, appropriate changes can be made.

PR #728, opened 2020-07-01
chore: create new jobs for React next

### LINKED ISSUE #726: Fix build for React next
Not sure exactly how/when this happened, but the build is failing: https://travis-ci.org/github/testing-library/react-testing-library/builds/703268660

Help welcome!

nickmccurdy:
**What**:

- Tests with both React stable and next for all builds (push and cron)
- Allow next releases to fail the build so we can still release
- Solves cron issues like #726 that are only caused by next releases

**Why**: #726 and greater visibility of what's failing in React next (before a cron build is started on schedule)

**How**: Split React next into a new section of the build matrix rather than just using it in cron builds

**Checklist**:

- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,14 @@ node_js:
   - 12
   - 14
   - node
+env:
+  - REACT_NEXT=false
+  - REACT_NEXT=true
 install:
   - npm install
   # as requested by the React team :)
   # https://reactjs.org/blog/2019/10/22/react-release-channels.html#using-the-next-channel-for-integration-testing
-  - if [ "$TRAVIS_EVENT_TYPE" = "cron" ]; then npm install react@next
+  - if [ "$REACT_NEXT" = true ]; then npm install react@next
     react-dom@next; fi
 script:
   - npm run validate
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Seems to be doing what I expected 😄 https://travis-ci.org/github/testing-library/react-testing-library/builds/704028047

Do we want to merge this and see if it fixes the next cron build tomorrow?

nickmccurdy:
I see what you mean. This technically does fix the overall build though, which is what the issue was originally for. Do you want to open another issue for the timing of React next?

eps1lon:
> This technically does fix the overall build though

As far as I can tell the test is still failing. Travis even marks the build as failed. This PR makes a failing build using `react@next` pass github checks.

nickmccurdy:
You're right. Would you still be comfortable with merging this so other issues also don't fail the build in the future? I'll keep #726 open to track the specific bug with React next that's currently failing.

nickmccurdy:
We merged this and kept #726 open to track the specific bug with React next.

PR #732, opened 2020-07-05
fix: React next fails on build (#726)

MatanBobi:
**What**: I've added a `scheduler-compat` module which will mock `unstable_scheduleCallback` for version of react that don't have the `Scheduler` package.

**Why**: Following https://github.com/testing-library/react-testing-library/issues/726, our build was failing for `react@next` so we needed to change the way we flush micro tasks.

**How**: I took `act-compat` as an example and took @kentcdodds code example [here](https://github.com/testing-library/react-testing-library/issues/726#issuecomment-653764644).
The only thing I added was an error handler because I had to suppress the error.

**Checklist**:

- [ ] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs) - Let me know if any documentation for this one is needed.
      
- [ ] Tests - I'm working on tests for this, it's failing in CI for Coverage.
- [ ] Ready to be merged
      

**This is a Draft PR just to get initial feedback, the code isn't good enough yet.**


src/flush-microtasks.js
```diff
@@ -17,6 +19,7 @@ function getIsUsingFakeTimers() {
 
 let didWarnAboutMessageChannel = false
 let enqueueTask
+
 try {
   // read require off the module object to get around the bundlers.
   // we don't want them to detect a require and bundle a Node polyfill.
```
kentcdodds, line 25: Let's add `Scheduler = nodeRequire.call(module, 'scheduler')` below the `enqueueTask` assignment below here (inside the `try` block).

Then, just after the `catch` block, add this:

```javascript
function scheduleCallback(cb) {
  const NormalPriority = Scheduler
    ? Scheduler.NormalPriority || Scheduler.unstable_NormalPriority
    : null

  const scheduleFn = Scheduler
    ? Scheduler.scheduleCallback || Scheduler.unstable_scheduleCallback
    : (cb) => cb();

  return isModernScheduleCallbackSupported
    ? scheduleCb(NormalPriority, cb)
    : scheduleCb(cb)
}
```

(sorry, github won't let me make a suggestion in a place that wasn't changed).

src/flush-microtasks.js
```diff
@@ -59,7 +62,11 @@ export default function flushMicroTasks() {
         jest.advanceTimersByTime(0)
         resolve()
       } else {
-        enqueueTask(resolve)
+        scheduleCallback(null, () => {
+          enqueueTask(() => {
+            resolve()
+          })
+        })
```
kentcdodds, line 69: With the change above, we no longer need this first argument

```suggestion
        scheduleCallback(() => enqueueTask(resolve))
```

src/scheduler-compat.js
```diff
@@ -0,0 +1,76 @@
...
+      return cb()
+    })
+  } else if (isModernScheduleCallbackSupported === false) {
+    return cb()
+  }
+
+  return Scheduler.unstable_scheduleCallback(NormalPriority, cb)
+}
+
+/* eslint no-console:0 */
```
kentcdodds, line 76: We don't need this file anymore.

```suggestion
```

src/flush-microtasks.js
```diff
@@ -1,3 +1,5 @@
+import scheduleCallback from './scheduler-compat'
```
kentcdodds, line 1: I don't think having the compat file is necessary.

```suggestion
import React from 'react'
import semver from 'semver'

const globalObj = typeof window === 'undefined' ? global : window

let Scheduler = globalObj.Scheduler

const isModernScheduleCallbackSupported = semver.satisfies(React.version, '>=16.8.6')
```

Also, let's move this code to below the comments, just above the `getIsUsingFakeTimers`

MatanBobi: You're definitely right about this one, I thought we don't want to use `semver`.. This was my go-to solution but since `act-compat` isn't using it and I didn't see it in the project, I didn't want to add it.
This simplifies everything.

MatanBobi:
> Thanks for this! What do you think of my suggestions?

I'm on it, sorry for making this PR too exhaustive, didn't think we'll be good with using `semver` :)

kentcdodds:
@all-contributors please add @MatanBobi for code

kentcdodds:
Thanks so much for your help! I've added you as a collaborator on the project (**with write access now**). Please make sure that you review the `other/MAINTAINING.md` and `CONTRIBUTING.md` files (specifically the bit about the commit messages and the git hooks) and familiarize yourself with the code of conduct (we're using [the contributor covenant](http://contributor-covenant.org/)). You might also want to [watch the repo](https://help.github.com/articles/watching-repositories/) to be notified when someone files an issue/PR. Please continue to make PRs as you feel the need (you can make your branches directly on the repo rather than your fork if you want). Thanks! And welcome to the team :)

MatanBobi:
Thanks a lot @kentcdodds!

PR #739, opened 2020-07-08
fix(flushMicroTasks): Fallback to no scheduler in case of exception

### LINKED ISSUE #736: scheduleFn is calling NormalPriority as a function rather than cb
As of https://github.com/testing-library/react-testing-library/pull/732/files we're seeing:

```
	TypeError: callback is not a function
	    at <Jasmine>
	    at scheduleFn (webpack:///node_modules/@testing-library/react/dist/@testing-library/react.esm.js:230:1 <- tests.webpack.js:27483:12)
	    at scheduleCallback (webpack:///node_modules/@testing-library/react/dist/@testing-library/react.esm.js:232:1 <- tests.webpack.js:27485:46)
	    at Object.then (webpack:///node_modules/@testing-library/react/dist/@testing-library/react.esm.js:245:1 <- tests.webpack.js:27498:9)
```

Looking at that PR it seems to have two sources of truth: isModernScheduleCallbackSupported based on the react version and NormalPriority based on the existence of Scheduler.

isModernScheduleCallbackSupported seems to be true yet Scheduler undefined.

Which causes it to call `scheduleFn(NormalPriority, cb)` rather than `scheduleFn(cb)`. And scheduleFn is `callback => callback()`.

MatanBobi:
Closes #736 

**What**: When we couldn't require the `scheduler` package we will now fallback to `isModernScheduleCallbackSupported = false` meaning we will use `callback => callback()`. 

**Why**: Following the conversation in #736 - We found this to be a proper solution for now in case the `Scheduler` wasn't loaded.
I still think we need to discuss this issue further and think if we can be able to figure a different way to require the `Scheduler` at runtime.

**How**: I've added an if statement in the catch block to see if `isModernScheduleCallbackSupported` is `true`, in that case I change it to `false` and write an error to console.

I think that we may need to separate the try catch to two different ones and handle each error for it's own. Update me what you think please :)
Also, within the error message we may want to link to the documentation stating how to fix this one.

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs) - We need to implement a documentation change for this one, please update me where you think it fits.
- [ ] Tests - N/A
- [ ] Ready to be merged

### DIFF
--- a/src/flush-microtasks.js
+++ b/src/flush-microtasks.js
@@ -20,9 +20,6 @@ function getIsUsingFakeTimers() {
 
 const globalObj = typeof window === 'undefined' ? global : window
 let Scheduler = globalObj.Scheduler
-let isModernScheduleCallbackSupported = satisfies(React.version, '>16.8.6', {
-  includePrerelease: true,
-})
 
 let didWarnAboutMessageChannel = false
 let enqueueTask
--- a/src/flush-microtasks.js
+++ b/src/flush-microtasks.js
@@ -20,7 +20,7 @@ function getIsUsingFakeTimers() {
 
 const globalObj = typeof window === 'undefined' ? global : window
 let Scheduler = globalObj.Scheduler
-const isModernScheduleCallbackSupported = satisfies(React.version, '>16.8.6', {
+let isModernScheduleCallbackSupported = satisfies(React.version, '>16.8.6', {
   includePrerelease: true,
 })
 
@@ -57,6 +57,17 @@ try {
           'Please file an issue at https://github.com/facebook/react/issues ' +
           'if you encounter this warning.',
       )
+    }
+
+    if (isModernScheduleCallbackSupported) {
+      // eslint-disable-next-line no-console
+      console.error(
+        'This environment does not support module requiring, so we cannot require the Scheduler. ' +
+          'To fix this you can require the Scheduler in your test setup file.',
+      )
+
+      // Fallback to no scheduleCallback Supported.
+      isModernScheduleCallbackSupported = false
     }
   }
 }
--- a/src/flush-microtasks.js
+++ b/src/flush-microtasks.js
@@ -59,18 +59,12 @@ try {
       )
     }
 
-    if (isModernScheduleCallbackSupported) {
-      // eslint-disable-next-line no-console
-      console.error(
-        'This environment does not support module requiring, so we cannot require the Scheduler. ' +
-          'To fix this you can require the Scheduler in your test setup file.',
-      )
-
-      // Fallback to no scheduleCallback Supported.
-      isModernScheduleCallbackSupported = false
-    }
   }
 }
+
+const isModernScheduleCallbackSupported = Scheduler && satisfies(React.version, '>16.8.6', {
+  includePrerelease: true,
+})
 
 function scheduleCallback(cb) {
   const NormalPriority = Scheduler


MatanBobi:
@kentcdodds - I created this one following the conversation in #736, I still think as @eps1lon pointed out that we should talk about a proper solution since this sounds more like a band-aid than a way to fix the problem :/

src/flush-microtasks.js
```diff
@@ -58,6 +58,17 @@ try {
...
+      console.error(
+        'This environment does not support module requiring, so we cannot require the Scheduler. ' +
+          'To fix this you can require the Scheduler in your test setup file.',
+      )
+
+      // Fallback to no scheduleCallback Supported.
+      isModernScheduleCallbackSupported = false
+    }
   }
 }
```
kentcdodds, line 73: It's not whether we can require the scheduler, but that there is no scheduler available. So I think we can make this change and that'll help resolve all the issues.

```suggestion

  }
}

const isModernScheduleCallbackSupported = Scheduler && satisfies(React.version, '>16.8.6', {
  includePrerelease: true,
})
```

src/flush-microtasks.js
```diff
@@ -20,7 +20,7 @@ function getIsUsingFakeTimers() {
 
 const globalObj = typeof window === 'undefined' ? global : window
 let Scheduler = globalObj.Scheduler
-const isModernScheduleCallbackSupported = satisfies(React.version, '>16.8.6', {
+let isModernScheduleCallbackSupported = satisfies(React.version, '>16.8.6', {
   includePrerelease: true,
 })
```
kentcdodds, line 25: With my suggested change below, we can remove these lines.

```suggestion
```

kentcdodds:
As for whether we should not use the scheduler at all, I'd love to avoid it as well, but I know for a fact that without these changes, we run into some serious issues when using fake timers. I spent two days banging my head against the wall and this is the solution.

I'd argue that `act` is an implementation detail as well, but we as the testing library, handle that for our users so they don't have to worry about it themselves. We do the same thing for cleanup. So yes, I agree that using the scheduler is an implementation detail of React, but it's our job to shield our users from implementation details. If we could avoid it that would be great, but considering we can't... 🤷‍♂️

src/flush-microtasks.js
MatanBobi: What do you mean when you say available? Since they use React 16.13.1, it should be available shouldn't it?
To me it looks like we couldn't require it using `nodeRequire.call(module, 'scheduler')`.

MatanBobi:
@kentcdodds I agree with the approach and I sure think that this change is required.
My only thought is that it looks like we're experiencing problems with bundling systems and us not being able to require the module..

kentcdodds:
Yeah, luckily *most* of the time it shouldn't matter. It only seems to matter in some edge cases. And in those cases, people using bundlers can simply set `window.Scheduler = require('scheduler')` and everything will work fine. I'm ok with that trade-off.

MatanBobi:
> Yeah, luckily _most_ of the time it shouldn't matter. It only seems to matter in some edge cases. And in those cases, people using bundlers can simply set `window.Scheduler = require('scheduler')` and everything will work fine. I'm ok with that trade-off.

Cool. Do you think we should add it in the docs/readme? 
That was my initial thought when I added the `console.error`, to reference the ones experiencing this issue to a fix.

kentcdodds:
The problem is for older versions of React which don't use/need the scheduler. I suppose we could add a little bit of logic to target folks who are using a version of React that uses the scheduler but we can't access it. If you want to do that, please do!

Issue #738, opened 2020-07-08
10.4.4 causing timeout issue with useEffect+jest.useFakeTimers


awesomeunleashed:
- `@testing-library/react` version: 10.4.4
- Testing Framework and version: Jest 24.9.0 (Create React App)
- DOM Environment: jsdom 14.1.0 (Create React App)

### Relevant code or config:
```js
// In a component
useEffect(() => {
  // any effect
}, [])

// In the tests
beforeEach(jest.useFakeTimers)
afterEach(jest.useRealTimers)
```

### What you did:

I have a component that has a `useEffect` in it and its tests use `jest.useFakeTimers`. My tests for the component passed until I updated to `@testing-library/react` 10.4.4.

### What happened:

The test now fails with this message:
```
Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.Error: Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.

      at mapper (node_modules/jest-jasmine2/build/queueRunner.js:25:45)
```

### Reproduction:

See https://github.com/awesomeunleashed/rtl-10-4-4-repro

### Problem description:

I believe these tests should still be passing.

### Suggested solution:

I imagine this has something to do with the scheduler changes in #732, as `useEffect` and `jest.useFakeTimers` sound like scheduling-related things, but I don't see anything obviously related without knowing the code.

kentcdodds:
This should be resolved with #739 which will be published in a moment. If it's not fixed in the next version, please comment here.

awesomeunleashed:
I updated my repro project to use 10.4.5, and I'm still seeing the same error.

kentcdodds:
Rats. If anyone would like to dig a little bit to find what the issue is that would be welcome :-)

marcosvega91:
As I can see from the code, the problem could be that `cleanup` is executed after `jest.useRealTimers`.

So when `microtasks` are flushed it will try to flush with `realTimers` instead of `fakeTimers`

MatanBobi:
> As I can see from the code, the problem could be that `cleanup` is executed after `jest.useRealTimers`.
> 
> So when `microtasks` are flushed it will try to flush with `realTimers` instead of `fakeTimers`.
> 
> If this is the case a possible solution, that i really don't like but it is my first thought, is to override `afterEach` function putting at the beginning the `cleanup`. In this way we are sure that `cleanup` is run before everything

[...]

PR #743, opened 2020-07-12
chore(flushMicroTasks): Use the mocked scheduler when available

MatanBobi:
Following a conversation with @eps1lon, he pointed out that maybe it would be better to mock the Scheduler (based on [this](https://github.com/facebook/react/pull/14964) PR by @acdlite).

This solution will handle both mocked version of scheduler and original version.

**What**: I've mocked the scheduler package in our setup-env and used the `flushAll` function which is available in the mocked scheduler package.

**Why**: React core team recommends to use a mocked version of the scheduler in test environments. We need to think if we want to recommend our users to mock the scheduler package also in their test environments. I think that if our users will mock the scheduler package, we won't need to know if someone is using fake timers or not and just call `flushAll`.

**How**: Checked for the presence of `unstable_flushAll` and used it if it exists. 

**Checklist**:

- [ ] Tests - N/A
- [ ] Ready to be merged
      

I'm opening this one to create the conversation here and see what you think about it.

### DIFF
--- a/src/flush-microtasks.js
+++ b/src/flush-microtasks.js
@@ -55,13 +55,14 @@ try {
           'if you encounter this warning.',
       )
     }
-
   }
 }
 
-const isModernScheduleCallbackSupported = Scheduler && satisfies(React.version, '>16.8.6', {
-  includePrerelease: true,
-})
+const isModernScheduleCallbackSupported =
+  Scheduler &&
+  satisfies(React.version, '>16.8.6', {
+    includePrerelease: true,
+  })
 
 function scheduleCallback(cb) {
   const NormalPriority = Scheduler
@@ -86,6 +87,9 @@ export default function flushMicroTasks() {
         // reproduce the problem, so there's no test for it. But it works!
         jest.advanceTimersByTime(0)
         resolve()
+      } else if (Scheduler && Scheduler.unstable_flushAll) {
+        Scheduler.unstable_flushAll()
+        enqueueTask(resolve)
       } else {
         scheduleCallback(() => enqueueTask(resolve))
       }
--- a/tests/setup-env.js
+++ b/tests/setup-env.js
@@ -1,1 +1,2 @@
 import '@testing-library/jest-dom/extend-expect'
+jest.mock('scheduler', () => jest.requireActual('scheduler/unstable_mock'))


src/flush-microtasks.js
```diff
@@ -86,6 +87,9 @@ export default function flushMicroTasks() {
         // reproduce the problem, so there's no test for it. But it works!
         jest.advanceTimersByTime(0)
         resolve()
+      } else if (Scheduler && Scheduler.unstable_flushAll) {
+        Scheduler.unstable_flushAll()
+        enqueueTask(resolve)
       } else {
         scheduleCallback(() => enqueueTask(resolve))
       }
```
kentcdodds, line 95:
```suggestion
      } else {
        Scheduler?.unstable_flushAll?.()
        enqueueTask(resolve)
      }
```

MatanBobi: I think that in that case we won't `scheduleCallback` in case we're not using the mocked scheduler, right?

MatanBobi:
Thanks @kentcdodds! 
Do you have any thoughts on whether we should recommend our users to mock the scheduler package or not?

src/flush-microtasks.js
```diff
@@ -86,6 +87,9 @@ export default function flushMicroTasks() {
         // reproduce the problem, so there's no test for it. But it works!
         jest.advanceTimersByTime(0)
         resolve()
+      } else if (Scheduler && Scheduler.unstable_flushAll) {
+        Scheduler.unstable_flushAll()
+        enqueueTask(resolve)
```
eps1lon, line 92: We don't have to queue the task. Should be able resolve immediately.

MatanBobi: I tried that, I wasn't able to understand why it didn't work and didn't get the time to look into it yesterday,
I'll look into it today.

eps1lon: Are we supposed to wait for microtasks in cleanup functions? Does it even make sense to run async code in the cleanup function?

kentcdodds: I'd we can avoid this being async that would be awesome, but I don't think we can avoid it because we can't assume people will be mocking the scheduler (especially when they're testing in environments with poor support for mocking).

I assumed that calling flushAll would mean we wouldn't need to worry about scheduling a callback. Sure wish we had a test that reproduces these timing issues. I have one in my bookshelf app that have me trouble before making all these changes. I can try these changes out there. I just couldn't simplify it to a reproducible example for inclusion in this project 😬

kentcdodds:
I think this is the right approach: https://github.com/testing-library/react-testing-library/pull/744

src/flush-microtasks.js
eps1lon: One suspicion I have is that you clean up your timers after the cleanup from testing-library runs. At least this is one issue I encountered in my mocha tests. Basically if you import from testing-libary we trigger an afterEach hook with cleanup and then in your test you call afterEach to cleanup timers. But this results in the cleanup of testing-libary being called with fake timers enabled which is fundamentally flawed (people can't expect to check for fake timers since they're not standardized).

If you can point me to the test in question I could check out if that is indeed the case.

MatanBobi:
> I think this is the right approach: #744

Yes, definitely. I'm closing this one. Thanks!

src/flush-microtasks.js
kentcdodds: The test that caused the problem was this one: https://github.com/kentcdodds/bookshelf/blob/85e2880abb3d3290c45a5a0ce061d3e8197b395a/src/__tests__/book-screen.js#L154-L175

But the test that failed was this one: https://github.com/kentcdodds/bookshelf/blob/85e2880abb3d3290c45a5a0ce061d3e8197b395a/src/__tests__/book-screen.js#L186-L194

I just tried to reproduce the problem and can't 😬 It's possible that I've changed the test enough that I don't end up in the edge case situation. It's also possible that my deps have changed to side-step the edge case as well. I'm just not certain.

Basically, the problem stems from scheduling some timers in the scheduler when fake timers are enabled, then switching back to real timers without advancing all of the timers for the scheduler's callbacks. This happens in the first test. Then when the next test runs it attempts to schedule callbacks which are never run. I *think* they're not run because the scheduler is still waiting for the first callbacks to run which never happens because they were scheduled with the fake timer.

eps1lon: > Basically, the problem stems from scheduling some timers in the scheduler when fake timers are enabled, then switching back to real timers without advancing all of the timers for the scheduler's callbacks.

I was about to ask why they don't automatically advance until all are finished but I guess it isn't that obvious what we mean by "finished" in the case of `setInterval`. Definitely need to look into it and start running at least all the timeouts before restoring fake timers if that isn't already the case.

kentcdodds: Yeah, part of me feels like this problem should be solved in either the scheduler or jest 🤔

eps1lon: "infinite timers" is also something I haven't considered: https://jestjs.io/docs/en/timer-mocks.

If we're recommending to cleanup between every test we should probably start recommending to clean up timers as well (`runOnlyPendingTimers`/`clearAllTimers` and `useRealTimers`) (and before testing-library cleanup):

- beforeEach: useFakeTimers
- test: render()
- afterEach: clearAllTimers
- afterEach: useRealTimers
- afterEach: cleanup

The problem right now is that you can't achieve that order with automatic cleanup.

This ensures that only the code you can "see" in your test uses fake timers and not unrelated cleanup code from third-party libraries that isn't aware of what kind of timers you're using.

kentcdodds: it seems to me like the best course of action would be to add a page in the documentation explaining the best way to use fake timers with testing library.

MatanBobi: > Basically, the problem stems from scheduling some timers in the scheduler when fake timers are enabled, then switching back to real timers without advancing all of the timers for the scheduler's callbacks. This happens in the first test. Then when the next test runs it attempts to schedule callbacks which are never run. I _think_ they're not run because the scheduler is still waiting for the first callbacks to run which never happens because they were scheduled with the fake timer.

@kentcdodds 
This was exactly the case in https://github.com/testing-library/react-testing-library/issues/738,
the scheduler used by react-dom in the test was using fake timers and since they weren't advanced, the next `scheduleCallback` we've added in the cleanup step was stuck and couldn't schedule the callback since a callback wasn't flushed before.

I agree that adding a page in the documentation to explain the usage of fake timers in testing library will be good. 
I don't mind working on this one if we think that's the best approach :)

kentcdodds: I think that's the best we can do, so let's do it 👍

MatanBobi: @kentcdodds, @eps1lon, I'm working on documenting this but having some second thoughts:
using `clearAllTimers` in `afterEach` may be problematic in some cases, for example if react's scheduler uses fake timers within our test and they aren't advanced, the callback won't be called and the booleans they count on won't be updated.. This might not be a problem though since it's at the cleanup stage.. 
Wdyt?

kentcdodds: Could we recommended `advanceAllTimers`?

eps1lon: > advanceAllTimers

Did you mean `runAllTimers`? We probably don't want to do this because [`runAllTimers` can cause infinite loops](https://jestjs.io/docs/en/timer-mocks#run-pending-timers). `runOnlyPendingTimers` seems like it fits our use case since this would run all scheduled timers from `scheduler`.

kentcdodds: 😅 yeah, that's what I meant 👍👍

MatanBobi: Thanks @eps1lon, I wasn't aware of the infinite loop that `runAllTimers` can cause. Using `runOnlyPendingTimers` sounds like a good approach but I'm not sure all testing frameworks have that ability (I think jasmine only has `tick` functionality but I've never really used it). Anyways, my examples will use jest.

eps1lon: > Using runOnlyPendingTimers sounds like a good approach but I'm not sure all testing frameworks have that ability (I think jasmine only has tick functionality but I've never really used it). Anyways, my examples will use jest.

Which is why I don't think we should handle this in testing-library. Or at at least provide and entry point that makes no assumption about the testing framework.

MatanBobi: > Which is why I don't think we should handle this in testing-library. Or at at least provide and entry point that makes no assumption about the testing framework.

Your right.. Our plan wasn't to handle this within testing-library, just explain what should be done by our users within a new documentation page.. Do you think we shouldn't do that either?

kentcdodds: Let's definitely do that 👍

PR #744, opened 2020-07-13
fix(cleanup): remove scheduler code from flush-microtasks

kentcdodds:
**What**: This removes the scheduler code from flush-microtasks

**Why**: Because the test that failed is fundamentally flawed

**How**:

1. Rewrite the test to illustrate why it's flawed (we can't and shouldn't guarantee that all scheduled tasks will run simply as a result of cleanup)
2. Rewrite the test to not test for that
3. Fix the implementation of flush-microtasks to not use the scheduler (basically revert everything we did for #726 and change the test instead).

**Checklist**:

~- [ ] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs)~ N/A
- [ ] Tests
~- [ ] Typescript definitions updated~ N/A
- [ ] Ready to be merged
      

Closes #743

### DIFF
--- a/src/__tests__/cleanup.js
+++ b/src/__tests__/cleanup.js
@@ -31,7 +31,7 @@ test('cleanup waits for queued microtasks during unmount sequence', async () => 
   const spy = jest.fn()
 
   const Test = () => {
-    React.useEffect(() => () => setImmediate(spy))
+    React.useEffect(() => () => setTimeout(spy, 200))
 
     return null
   }
--- a/src/__tests__/cleanup.js
+++ b/src/__tests__/cleanup.js
@@ -27,11 +27,11 @@ test('cleanup does not error when an element is not a child', async () => {
   await cleanup()
 })
 
-test('cleanup waits for queued microtasks during unmount sequence', async () => {
+test('cleanup runs effect cleanup functions', async () => {
   const spy = jest.fn()
 
   const Test = () => {
-    React.useEffect(() => () => setTimeout(spy, 200))
+    React.useEffect(() => spy)
 
     return null
   }
--- a/package.json
+++ b/package.json
@@ -44,8 +44,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.10.3",
-    "@testing-library/dom": "^7.17.1",
-    "semver": "^7.3.2"
+    "@testing-library/dom": "^7.17.1"
   },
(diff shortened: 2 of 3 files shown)


src/__tests__/cleanup.js
```diff
@@ -31,7 +31,7 @@ test('cleanup waits for queued microtasks during unmount sequence', async () =>
   const spy = jest.fn()
 
   const Test = () => {
-    React.useEffect(() => () => setImmediate(spy))
+    React.useEffect(() => () => setTimeout(spy, 200))
```
kentcdodds: Users can do all kinds of async things as a result of their cleanup functions and it's not reasonable for them to expect that we'll wait until all those async things are done. They'll have to wait for those things themselves.

So we shouldn't have a `setImmediate` in here at all.

src/__tests__/cleanup.js
```diff
@@ -27,11 +27,11 @@ test('cleanup does not error when an element is not a child', async () => {
   await cleanup()
 })
 
-test('cleanup waits for queued microtasks during unmount sequence', async () => {
+test('cleanup runs effect cleanup functions', async () => {
   const spy = jest.fn()
 
   const Test = () => {
-    React.useEffect(() => () => setImmediate(spy))
+    React.useEffect(() => spy)
```
kentcdodds, line 34: All we can guarantee is that the cleanup function is called, so that's all we'll test.

eps1lon: :+1: The problem is that we can't know what kind of task is queued from the cleanup function. We only supported `setImmediate` but any longer running task wasn't caught as far as I can tell.

kentcdodds:
Thanks for all the work you did anyway @MatanBobi!

MatanBobi:
Glad I was able to help @kentcdodds

pke:
@kentcdodds I think the react-native package suffers from a similar bug. When I add `jest.useFakeTimers("modern")` all tests suddenly time out.

MatanBobi:
> @kentcdodds I think the react-native package suffers from a similar bug. When I add `jest.useFakeTimers("modern")` all tests suddenly time out.

Hi @pke!
Did you happen to look at our [using fake timers page](https://testing-library.com/docs/using-fake-timers)?
Are you setting back to real timers in the `afterEach` hook?
A code example would be really helpful :)

PR #746, opened 2020-07-15
fix(unmount): Flush useEffect cleanup functions syncronously

eps1lon:
**What**:

Flush cleanup functions from `useEffect` synchronously on `unmount`.

**Why**:

`react@next` no longer flushes cleanup functions from `useEffect` synchronously.
Verified in https://travis-ci.org/github/testing-library/react-testing-library/builds/708314737

**How**:

Wrap `unmountComponentAtNode` in `act`.

**Checklist**:

- [x] Tests
- [x] Ready to be merged
      

I'm not touching `cleanup` and `flush-microtasks` for now. We can do that in a follow-up PR that we test extensively against existing repos.

### DIFF
--- a/src/__tests__/render.js
+++ b/src/__tests__/render.js
@@ -87,3 +87,17 @@ test('renders options.wrapper around node', () => {
 </div>
 `)
 })
+
+test('flushes useEffect cleanup functions sync on unmount()', () => {
+  const spy = jest.fn()
+  function Component() {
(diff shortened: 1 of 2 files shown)


marcosvega91:
just a question, this should be done also here? https://github.com/testing-library/react-testing-library/blob/240900c3d0bf72a94d3c0dfc85a5214ef3b16f63/src/pure.js#L109

PR #768, opened 2020-08-21
feat: Use act to flush instead of custom implementation

eps1lon:
Replaces #514 with wrapping `unmountComponentAtNode` in `act` i.e. we no longer flush microtasks manually. We need to wrap it in `act` for React 17 so that effect cleanup functions are flushed.

I extracted the original description into a test (as I understood the description) and the test is still passing with this chamge

The current implementation is problematic since it relies on a specific implementation of the scheduler i.e. a specific config. Some aspects already changed over time and might change again (considering main thread scheduling proposal for chrome). So having the custom implementation is a liability that already caused problems in the past.

As far as I can tell the underlying problem is tests that do not properly reset their timers. You shouldn't use fake timers between tests if you're using third party libraries. As soon as they use any async APIs in beforeEach or afterEach (like we're doing) you can get into incredibly hard to debug situations. At least we shouldn't be solving this problems since jest's fake timers aren't the only fake timers. 

We also currently didn't handle any macrotasks so I don't even think we solved that many issues.

PS: Not having any repros (does not have to be deterministic) is problematic for code that is supposed to stay. I'd release this in a major or alpha to see if this causes any problems. My guess is that we should add a warning in our automatic cleanup if fake timers are still on.


eps1lon:
https://github.com/testing-library/react-testing-library/pull/768#issuecomment-678702395

Yep, that's what I think is the correct pattern for using fake timers. Otherwise you're creating an environment for a test that leaks to possibly unrelated files. 

Though I think we should add a warning to our `cleanup` function if we detect fake timers when cleaning up. Then we can explain the reason and it hopefully helps people detect these issues.

And to re-iterate: This change is backwards incompatible. I suspect most people don't clean up their timers so we should probably release this as a major.

kentcdodds:
Yep, makes sense.

In my own tests, I'm thinking that I'll do this in my setup file:

```javascript
afterEach(() => jest.useRealTimers())
```

And then just call `jest.useFakeTimers()` for the ones that need fake timers. However, React isn't the only thing that has timers that may need to be run, so I think we should probably still recommend running pending timers:

```javascript
afterEach(() => {
  jest.runOnlyPendingTimers()
  jest.useRealTimers()
})
```

So not a ton will change for the way *I* do things (though I'm sure it will for others). In any case, I'm super jazzed about this change. Thanks!

kentcdodds:
Let's think about other breaking changes we could make at this point as well...

src/__tests__/cleanup.js
```diff
@@ -37,6 +37,87 @@ test('cleanup runs effect cleanup functions', async () => {
...
-  await cleanup()
+  cleanup()
   expect(spy).toHaveBeenCalledTimes(1)
 })
+
+describe('fake timers and missing act warnings', () => {
+  beforeEach(() => {
+    jest.resetAllMocks()
+    jest.spyOn(console, 'error').mockImplementation(() => {
+      // assert messages aexplicitly
```
eps1lon, line 48:
```suggestion
      // assert messages explicitly
```

kentcdodds:
Found the problem was with react-query. This change makes things better with that as well as it turns out.

https://github.com/tannerlinsley/react-query/pull/909#issuecomment-683178702

Unless anyone can think of other breaking changes we should consider, I think this is ready to go.

eps1lon:
@kentcdodds Yeah I only had this as a draft because I was hoping for some more tests in other repos. With bookshelf tested this looks fine by me. 

We had some pretty bad timing issues with React 17 in Material-UI and and already use patch internally.

Without this patch we'll probably see some very odd bug reports for React 17 depending on the test runner/timers used. I just hope we don't swallow some missing act warnings but since we're not inside the test this seems safe to do.

kentcdodds:
Let's do a beta release with this.

kentcdodds:
Another thing to consider is whether we want to make any breaking changes to DOM Testing Library which would impact this lib and we'd like to bundle up...

PR #782, opened 2020-09-08
doc: update the cheat sheet pdf + source file to fix typo

tapico-weyert:
**What**:
Fixes a typo in the Javascript in the Cheat sheet pdf

**Why**:
🐶

**How**:
I have updated the Affinity Publisher document and exported a PDF 

**Checklist**:
N/A


kentcdodds:
Thanks! It looks like the afpub doc was deleted 🤔

weyert:
Hmm, that wasn't the intention and the appears to be still in my repo directory. Did get a lot smaller by the looks of it

kentcdodds (approved):
Ah, that makes sense. Thanks!

PR #806, opened 2020-10-26
docs: update complex example

lidoravitan:
**What**: 
I update README.md complex example section. So, the test will be valid.
**Why**:
1) There was missing variable. 
2) I have seen that the promise always succeeds, so the tests always fail. The Promise returned from fetch() won’t reject on HTTP error status even if the response is an HTTP 404 or 500([reference](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch)).

**How**:
I added the missing variable and fixed the promise to reject in case of `200<` status code.

**Checklist**:

- [ ] Tests - N/A
- [ ] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,7 +233,9 @@ function Login() {
           password: passwordInput.value,
         }),
       })
-      .then(r => r.json())
+      .then((r) =>
+        r.json().then((data) => (r.ok ? data : Promise.reject(data)))
+      )
       .then(
         user => {
           setState({loading: false, resolved: true, error: null})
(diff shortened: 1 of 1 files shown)


MatanBobi (approved):
Looks good to me :)

kentcdodds:
@all-contributors please add @lidoravitan for docs

MatanBobi:
@kentcdodds It looks like the build failed for `node` in `typecheck` step, but it doesn't seem related to this PR.. Any ideas?

kentcdodds:
Node 15 messed things up with npx: https://github.com/npm/cli/issues/1935

I think we should just get rid of `node` from the travis config personally.

MatanBobi:
> Node 15 messed things up with npx: [npm/cli#1935](https://github.com/npm/cli/issues/1935)
> 
> I think we should just get rid of `node` from the travis config personally.

I can take that if you think we should do it.
But this will leave us with `10.14.2`, `12` and `14`, is that ok?

kentcdodds:
Yeah. I'm not sure why we'd need to test on node 15 🤷‍♂️

kentcdodds:
I mean, I understand that it could help us address breaking changes in the future, but it's unlikely that will be a problem for us to solve anyway.

kentcdodds:
What would be even better is if a failure of `node` does not trigger a build failure but rather could just be a warning of some kind.

PR #808, opened 2020-10-27
build: allow node latest build to fail

MatanBobi:
This is due to a braking change in npx adding a -y flag

**What**: Allow node latest build to fail because of a [change](https://github.com/npm/cli/issues/1935) in npx requiring us to add a `-y` flag to our CI.

**Why**: At the moment, our build is failing for Node 15.

**How**: Add the node latest version to the `allow_failures` section.

**Checklist**:

- [ ] Tests - N/A
- [ ] Ready to be merged
      

This was raised in this [comment](https://github.com/testing-library/react-testing-library/pull/806#issuecomment-716793552) by @kentcdodds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ jobs:
   allow_failures:
     - REACT_DIST=next
     - REACT_DIST=experimental
+    - node_js: node
   include:
     - stage: release
       node_js: 14


eps1lon (changes requested):
But now any failure in node would be displayed as green.

Does travis not allow conditional scripts based on variables? We should rather adjust the jobs accordingly i.e. run `npx -y codecov@3` in node 15

MatanBobi:
> But now any failure in node would be displayed as green.
> 
> Does travis not allow conditional scripts based on variables? We should rather adjust the jobs accordingly i.e. run `npx -y codecov@3` in node 15
> 
> Edit: just found
> 
> > npm_config_yes=true npx mocha
> 
> -- [npm/cli#1935 (comment)](https://github.com/npm/cli/issues/1935#issuecomment-706278238)

The allow_failures is only for `node` and not for the specific versions if I understood correctly as you can see [here](https://travis-ci.org/github/testing-library/react-testing-library/builds/739221760)..

But I do agree that it would be better to fix it specifically for node 15. I'll have a look as I'm not 100% familiar with travis.

eps1lon:
>  I'll have a look as I'm not 100% familiar with travis.

I would start with `npm_config_yes=true npx mocha` first. Seems like the easier choice.

kentcdodds:
We could definitely go with that approach, but I'm not a big fan of experimental software (odd versions of node) blocking our ability to release 😬 If something else comes up that would be pretty annoying.

eps1lon:
> We could definitely go with that approach, but I'm not a big fan of experimental software (odd versions of node) blocking our ability to release  If something else comes up that would be pretty annoying.

I'm hesitant as well but `npm_config_yes=true npx -y codecov@3` looks not that invasive to me.

kentcdodds:
I agree it's not a big deal. What I mean is considering Node 15 is experimental software, our build could break for confusing reasons and if that stops us from shipping that would be annoying since we technically don't support experimental versions of node.

kentcdodds:
I guess I'm saying we can add that to get around this issue, but I *also* want to make it so the node 15 build doesn't fail CI.

eps1lon:
Why do we have the `node` target in travis to begin with? We already list 14 and 16 is 6mo out. Would it make more sense to replace `node` with `lts/*`? Otherwise I don't understand what the point is to test on node 15. Do we support it or not?

Also want to be clear that node 15 is not experimental. It's as stable as any released software. It just is not considered LTS.

kentcdodds:
Yeah, I think it would make sense to remove `node` from the list so we don't have unexpected surprises in the future. Technically we do support node 15 thanks to our `>=10` listing in the package.json.

MatanBobi:
I'm closing this one since @eps1lon opened https://github.com/testing-library/react-testing-library/pull/809

Thanks everyone :)

PR #809, opened 2020-10-27
test: Fix node 15

eps1lon:
**What**:

Explicitly add node 15 to the matrix and fix compat issues.

**Why**:

We declare that we support node >= 10 so we should test accordingly. Now that node 15 is released we can declare it in the matrix.

**How**:

Follow advise in https://github.com/npm/cli/issues/1935 until npm fixes its CI detection.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,14 @@ node_js:
   - 10.14.2
   - 12
   - 14
+  - 15
+env:
+  - REACT_DIST=latest
+  - REACT_DIST=next
+  - REACT_DIST=experimental
+  # node15 compat
+  # Revisit once https://github.com/npm/cli/pull/2047 is released.
+  - npm_config_yes=true
 env:
   - REACT_DIST=latest
   - REACT_DIST=next
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,15 @@ node_js:
   - 10.14.2
   - 12
   - 14
+  - 15
   - node
 env:
   - REACT_DIST=latest
   - REACT_DIST=next
   - REACT_DIST=experimental
+  # node15 compat
+  # Revisit once https://github.com/npm/cli/pull/2047 is released.
+  - npm_config_yes=true
 install:
   - npm install
   # as requested by the React team :)
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ env:
   - REACT_DIST=next
   - REACT_DIST=experimental
 install:
+  # 7.0.6 fixed CI environments prompting user input.
+  # Can be removed once node15 bumps the shipped npm version.
+  - npm install --global npm@>=7.0.6
   - npm install
   # as requested by the React team :)
   # https://reactjs.org/blog/2019/10/22/react-release-channels.html#using-the-next-channel-for-integration-testing
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,15 +7,10 @@ node_js:
   - 10.14.2
   - 12
   - 14
-  - 15
-  - node
 env:
   - REACT_DIST=latest
   - REACT_DIST=next
   - REACT_DIST=experimental
-  # node15 compat
-  # Revisit once https://github.com/npm/cli/pull/2047 is released.
-  - npm_config_yes=true
 install:
   - npm install
   # as requested by the React team :)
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,11 +7,15 @@ node_js:
...
   - 14
+  - 15
   - node
 env:
   - REACT_DIST=latest
   - REACT_DIST=next
   - REACT_DIST=experimental
+  # node15 compat
+  # Revisit once https://github.com/npm/cli/pull/2047 is released.
+  - npm_config_yes=true
```
MatanBobi, line 18: Is there any downside to putting it here rather than adding it only for node 15? That was what I was initially thinking..

MatanBobi (approved):

.travis.yml
eps1lon: They said specifically that it is backwards compatible. I do set it conditionally but why bloat the config if it works with every version?

eps1lon:
@kentcdodds Could you elaborate why we have the separate `node` identifier? As far as I can tell travis  is not deduplicating the matrix so you end up with two runs for node 15 (before its release we had two runs for node 14): https://travis-ci.org/github/testing-library/react-testing-library/builds/739343261. It makes more sense to me to remove the `node` entry and update the config once node 16 is released.

`node` was added in https://github.com/testing-library/react-testing-library/commit/ea67b170fccfaf15b799d5aff5579f638b095892

MatanBobi:
> @kentcdodds Could you elaborate why we have the separate `node` identifier? As far as I can tell travis is not deduplicating the matrix so you end up with two runs for node 15 (before its release we had two runs for node 14): https://travis-ci.org/github/testing-library/react-testing-library/builds/739343261. It makes more sense to me to remove the `node` entry and update the config once node 16 is released.
> 
> `node` was added in [ea67b17](https://github.com/testing-library/react-testing-library/commit/ea67b170fccfaf15b799d5aff5579f638b095892)

@eps1lon it looks like @kentcdodds wrote in https://github.com/testing-library/react-testing-library/pull/808#issuecomment-717353821 that we can remove the separate `node` identifier.. So I guess you can go ahead and remove it :)

.travis.yml
MatanBobi: It looks [here](https://github.com/npm/cli/issues/1935#issuecomment-706278238) that it's compatible for npm 6 and 7, node 10.2.1 still works with npm 5, but we're running this only for node 10.14.2 so this will pass.. We already wrote there that "technically we support 10.0.0, but some of our tooling doesn't" so we're covered there.
Thanks!

nickmccurdy:
I thought we still wanted to test Node 15, but using `15` instead of `node` so the version doesn't change over time

eps1lon:
I can do this all day :grimacing:

.travis.yml
```diff
@@ -7,7 +7,14 @@ node_js:
...
   - 14
-  - node
+  - 15
+env:
+  - REACT_DIST=latest
+  - REACT_DIST=next
+  - REACT_DIST=experimental
+  # node15 compat
+  # Revisit once https://github.com/npm/cli/pull/2047 is released.
+  - npm_config_yes=true
```
kentcdodds, line 17: I think we can ditch this: https://github.com/npm/cli/issues/1935#issuecomment-717562445

nickmccurdy: We might have to manually install that version of npm though (such as with `npm install --global npm@7`). Also the `env` section is duplicated, we only need one.

nickmccurdy (approved):

nickmccurdy:
Looks ready to merge for me

.travis.yml
```diff
@@ -7,12 +7,15 @@ node_js:
...
-  - node
+  - 15
 env:
   - REACT_DIST=latest
   - REACT_DIST=next
   - REACT_DIST=experimental
 install:
+  # 7.0.6 fixed CI environments prompting user input.
+  # Can be removed once node15 bumps the shipped npm version.
+  - npm install --global npm@>=7.0.6
```
ljharb, line 18: Try this instead, in general:
```suggestion
before:install:
  - nvm install-latest-npm
install:
```

That will always use the latest working npm version on whatever node version you're using.

kentcdodds: Thank you @ljharb!

PR #811, opened 2020-10-29
chore: update React imports

MichaelDeBoey:
https://twitter.com/kentcdodds/status/1321885232878923776

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ afterAll(() => {
 
 ```jsx
 // hidden-message.js
-import React from 'react'
+import * as React from 'react'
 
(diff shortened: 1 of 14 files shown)


kentcdodds:
Oh, also, thanks!

eps1lon:
> I don't think this should make any substantive difference to consumers right?

Might be better when bundled with webpack but slightly worse when transpiled to commonJS. Though ultimately negligible. If we see actual issues `const React = require('react')`  is the best thing for modules not published as ES modules.

MichaelDeBoey:
This won't make any difference indeed.

PR #812, opened 2020-10-29
chore: adjust .travis to install latest npm

kentcdodds:
As recommended by @ljharb https://github.com/testing-library/react-testing-library/pull/809/files#r514441103

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,9 @@ env:
   - REACT_DIST=latest
   - REACT_DIST=next
   - REACT_DIST=experimental
+before_install:
+  - nvm install-latest-npm
 install:
-  # 7.0.6 fixed CI environments prompting user input.
-  # Can be removed once node15 bumps the shipped npm version.
-  - npm install --global npm@>=7.0.6
   - npm install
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Out of curiosity, are we still supporting Node versions that don't support npm 7?

ljharb (approved):
No, but this way you don’t have to keep track of which npm supports which node.

kentcdodds:
Thanks @ljharb!

@All-contributors please add @ljharb for review

nickmccurdy:
FYI: You can use the `Add Suggestion` button on GitHub (or `git commit --author "..."` to commit changes manually) to commit something authored by someone else. git/GitHub will display both users on the commit.

nickmccurdy:
@kentcdodds This installs npm 6, not 7. Is that intentional or would you like me to PR to use the latest working version of 7?

ljharb:
That’s because npm 7 isn’t marked latest yet.

nickmccurdy:
I'm aware. I wasn't sure if Kent wanted to specifically use 6 or 7, or just have the fix for #809.

PR #825, opened 2020-11-14
Revert "chore: adjust .travis to install latest npm"

nickmccurdy:
Reverts testing-library/react-testing-library#812, which installs npm 6 not 7


kentcdodds (commented):
I thought we'd already changed this to GitHub actions. We should just do that.

PR #826, opened 2020-11-14
chore: switch to github actions

nickmccurdy:
Generated with `yo kcd-oss`

See https://github.com/testing-library/dom-testing-library/pull/809 for rationale and history


nickmccurdy:
Any ideas why this isn't running on GitHub Actions but testing-library/dom-testing-library#809 was? I was hoping to test this before merging it.

MatanBobi:
> Any ideas why this isn't running on GitHub Actions but [testing-library/dom-testing-library#809](https://github.com/testing-library/dom-testing-library/pull/809) was? I was hoping to test this before merging it.

I think that the workflow folder should be under the `.github` folder so you probably missed the `.` here, I think it will fix that :)

nickmccurdy:
Uh oh... I think that needs to be fixed upstream in `kcd-oss`

nickmccurdy:
TypeScript ESLint issues are failing the build

MatanBobi:
> TypeScript ESLint issues are failing the build

Looks like this is because of the `kcd-scripts` version that needs to be upgraded like @kentcdodds did [here](https://github.com/testing-library/dom-testing-library/pull/817) :)

nickmccurdy:
I keep getting this error after upgrading it, adding a `tsconfig.json` does not help:
```json
[lint] Error: Error while loading rule '@typescript-eslint/no-unsafe-assignment': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
```

kentcdodds:
I'm working on fixing this now

kentcdodds:
Alrighty, now eslint-config-kentcdodds will automatically find the `tsconfig.json` and set that as the `project`. This passes locally so we should be good here as well.

kentcdodds (approved):
Super duper! Thanks 👏

types/test.tsx
```diff
@@ -57,14 +59,22 @@ async function testDebug() {
...
   await waitFor(() => {})
 }
+
+/*
+eslint
+  testing-library/prefer-explicit-assert: "off",
+  testing-library/no-wait-for-empty-callback: "off",
+  testing-library/no-debug: "off",
+  testing-library/prefer-screen-queries: "off"
+*/
```
nickmccurdy, line 80: Could we disable these globally or do they need to be enabled in other tests?

kentcdodds: We only need them in this test. So I don't see the point in doing that, but if we did create a separate file within this directory that needed these rules disabled then we could add these to an eslintrc file within this directory.

Issue #830, opened 2020-11-18
React Testing Library API types prevent usage of SVG elements


JoeBobMiles:
- `@testing-library/react` version: 11.1.1
- Testing Framework and version: `jest` 25.1.0
- DOM Environment: `jsdom`

### Relevant code or config:

```tsx
import * as React from "react";
import { render } from "@testing-library/react";
import "@testing-library/jest-dom/extend-expect";

test("Is a circle", () => {
  const { container } = render(
    <circle cx=0 cy=0 />,
    {
      container:
        document.body.appendChild(
          document.createElementNS("http://www.w3.org/2000/svg", "svg")),
    });

  expect(container.firstChild.nodeName).toBe("circle");
});
```

### What you did:

Trying to set SVG as the container element for testing SVG React components.

### What happened:

Typescript threw a type error stating that `render()` requires an HTMLElement.

### Reproduction:

I created a [repository](https://github.com/JoeBobMiles/react-testing-library-svg-api-bug) illustrating my problem.

### Problem description:

React allows components to use HTML and SVG tags. SVG tags do not work properly in an HTML context, and consequently must be rendered inside an SVG context. This can be only done by using an SVG tag as the containing element. Since the API's types disallow this, users have to wrap all their SVG based components in SVG tags manually and introducing unnecessary nesting to the render output.

### Suggested solution:

Propose the following patch to `@testing-library/react/types/index.d.ts`:
```ts
// TypeScript Version: 3.8

import {OptionsReceived as PrettyFormatOptions} from 'pretty-format'
import {queries, Queries, BoundFunction} from '@testing-library/dom'
import {act as reactAct} from 'react-dom/test-utils'

export * from '@testing-library/dom'

type DOMElement = HTMLElement | SVGElement;

export type RenderResult<Q extends Queries = typeof queries> = {
  container: DOMElement
  baseElement: DOMElement
  debug: (
    baseElement?:
      | DOMElement
      | DocumentFragment
      | Array<DOMElement | DocumentFragment>,
    maxLength?: number,
    options?: PrettyFormatOptions,
  ) => void
  rerender: (ui: React.ReactElement) => void
  unmount: () => boolean
  asFragment: () => DocumentFragment
} & {[P in keyof Q]: BoundFunction<Q[P]>}

export interface RenderOptions<Q extends Queries = typeof queries> {
  container?: DOMElement
  baseElement?: DOMElement
  hydrate?: boolean
  queries?: Q
[...]

PR #833, opened 2020-11-19
fix(types): Allow all HTML and SVG elements in render.

JoeBobMiles:
**What**:

Types disallowed SVG elements, see #830.

**Why**:

To allow both HTML and SVG elements to be passed to React Testing Library functions without throwing type errors in TypeScript.

**How**:

Exchanged `HTMLElement` type for the broader `Element` type that encompasses all elements that can appear in the DOM, not just HTML.

**Checklist**:

- [ ] Tests N/A
- [x] Ready to be merged


nickmccurdy:
I've noticed there are also types for `Node` (less specific) and `HTMLOrSVGElement` (more specific). Could either of those be better?

eps1lon (approved):
> I've noticed there are also types for Node (less specific) and HTMLOrSVGElement (more specific). Could either of those be better?

`Element` is the exact interface we need because that's what `react-dom` supports. Practically `HTMLOrSVGElement` works because only `HTMLElement` and `SVGElement` extend `Element`. But we don't need any properties from `HTMLOrSVGElement` so we should stick with the correct type.

`Node` would allow non-sense like `Attr` nodes or `Text` nodes.

eps1lon:
Though there could be some fallout if we're now to generic for practical purposes. We'll see in the issue reports if we should add a type parameter for the container or if `HTMLOrSVGElement` is sufficient.

PR #834, opened 2020-11-19
chore(types): add test for rendering SVG elements

nickmccurdy:
**What**: Adding regression tests for #830 (implemented in #833)

**Why**: It helps to have a regression test for the original issue when adding a fix

**How**: TypeScript type test based on sample code in #830

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/types/test.tsx
+++ b/types/test.tsx
@@ -44,6 +44,15 @@ export function testRenderOptions() {
   render(<div />, options)
 }
 
+export function testSVGRenderOptions() {
+  const container = document.createElementNS(
+    'http://www.w3.org/2000/svg',
+    'svg',
+  )
+  const options = {container}
+  render(<svg />, options)
+}
+
 export function testFireEvent() {
   const {container} = render(<button />)
   fireEvent.click(container)


nickmccurdy:
I wonder if there's a way to get TypeScript test coverage using a logical specification DSL

kentcdodds (approved):
Thanks!

just-boris:
Why was this released as a minor version bump?

This breaks the following code:

```tsx
function findSomething(element: HTMLElement) {...}


const {container} = render(...);

findSomething(container);
//            ~~~~~~~~~
// Argument of type 'Element' is not assignable to parameter of type 'HTMLElement'.
//  Type 'Element' is missing the following properties from type 'HTMLElement': accessKey, accessKeyLabel, autocapitalize, // dir, and 107 more.(2345)
```

This is a breaking change, and should have waited until the major version 12

eps1lon:
> This is a breaking change, and should have waited until the major version

Sorry to hear you're having trouble with the latest patch.

You probably mean #833. The PR you commented on did not affect published code.

Note that any change is potentially backwards incompatible. Especially type related changes are very likely to break some code. This change in particular was bug fix. If your code relied on a bug then the change is backwards incompatible for you. Note that any `@types/*` package is already pushing type changes in patch versions for this exact reason.

However, we cannot always make allowances in these instances. Otherwise we would need to release any change with a major version.

just-boris:
Oh correct, I mixed these two PRs up.

I assumed that the signature of the public API remains unchanged. E.g. if `render()` function returns a certain type, it will continue returning it until the next major release.

Do you have any suggestions on how to protect against unexpected changes? The pattern is very common, there is a [real example (last line in the file)](https://unpkg.com/browse/@awsui/components-react@3.0.22/test-utils/dom/index.d.ts) which made me respond to this PR.

eps1lon:
> Do you have any suggestions on how to protect against unexpected changes?

Working on a fix right now.

eps1lon:
@just-boris Subscribe to https://github.com/testing-library/react-testing-library/pull/868 for further updates.

PR #851, opened 2020-12-11
[render] Add support for async side effects on mount

### LINKED ISSUE #850: Add handling of async side effects to render function

Add support for wrapping async render events in `act`.

By default, async side effects dispatched at mount aren't wrapped in act. It would be nice if RTL could handle this.

### Suggested implementation:

Before:

```tsx
it('some test', async () => {
  const myThing = render(<SomeComponent />);
  await act(async () => {}); // <- wait for async action on mount
});
```

After:

```tsx
  const myThing = await render(<SomeComponent />);  // <- wait for async action on mount
```

Could provide an option to more strongly differentiate between sync and async renders (not preferred)

```tsx
  const myThing = await render(<SomeComponent />, { async: true });
```

### Teachability, Documentation, Adoption, Migration Strategy:

Created a [PR Here](https://github.com/testing-library/react-testing-library/pull/851)

andyrichardson:
**What**: Fix #850

**Why**: See #850

**How**: See code

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/pure.js
+++ b/src/pure.js
@@ -96,6 +96,10 @@ function render(
       }
     },
     ...getQueriesForElement(baseElement, queries),
+    then: async (callback) => {
+      await act(async () => {});
+      callback();
+    }
   }
 }
 
--- a/src/pure.js
+++ b/src/pure.js
@@ -100,10 +100,7 @@ function render(
   
   return {
     ...response,
-    then: async (callback) => {
-      await act(async () => {});
-      callback(response);
-    }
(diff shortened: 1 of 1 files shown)


nickmccurdy:
As explained in https://github.com/testing-library/react-testing-library/issues/850#issuecomment-743237370, this won't necessarily pass for async effects/lifecycles as they may happen after the next tick. It would be better to explicitly await something that doesn't rely on implementation details, like `await findByText("Done loading")`

PR #868, opened 2021-01-26
fix(render): Default to HTMLElement in returned container

eps1lon:
**What**:

Fixes https://github.com/testing-library/react-testing-library/pull/834#issuecomment-767796456

**Why**:

By default we create a `div` but declare that the type is `Element`. This is not specific enough for most use cases since e.g. `offsetHeight` isn't implemented in `Element`

**How**:

Adding a type parameter as anticipated in https://github.com/testing-library/react-testing-library/pull/833#issuecomment-730280641

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/types/test.tsx
+++ b/types/test.tsx
@@ -3,7 +3,7 @@ import {render, fireEvent, screen, waitFor} from '.'
 import * as pure from './pure'
 
 export async function testRender() {
-  const page = render(<div />)
+  const page = render(<button />)
 
   // single queries
   page.getByText('foo')
@@ -22,7 +22,7 @@ export async function testRender() {
 }
 
 export async function testPureRender() {
-  const page = pure.render(<div />)
+  const page = pure.render(<button />)
 
   // single queries
   page.getByText('foo')
@@ -43,8 +43,9 @@ export async function testPureRender() {
 export function testRenderOptions() {
   const container = document.createElement('div')
   const options = {container}
-  const {container: returnedContainer} = render(<div />, options)
-  // Unclear why TypeScript infers `HTMLElement` here when the the input `container` is `HTMLDivElement`
+  const {container: returnedContainer} = render(<button />, options)
+  // Unclear why TypeScript infers `HTMLElement` here when the the input `container` is `HTMLDivElement`.
+  // It's working for `testSVGRenderOptions`. 
   // Hopefully this breaks someday and we can switch to
   // expectType<HTMLDivElement, typeof returnedContainer>(returnedContainer)
   expectType<HTMLElement, typeof returnedContainer>(returnedContainer)
@@ -56,7 +57,7 @@ export function testSVGRenderOptions() {
     'svg',
   )
   const options = {container}
-  const {container: returnedContainer} = render(<svg />, options)
+  const {container: returnedContainer} = render(<path />, options)
   expectType<SVGSVGElement, typeof returnedContainer>(returnedContainer)
 }
 
--- a/types/test.tsx
+++ b/types/test.tsx
@@ -17,6 +17,7 @@ export async function testRender() {
 
   // helpers
   const {container, rerender, debug} = page
+  expectType<HTMLElement, typeof container>(container)
   return {container, rerender, debug}
 }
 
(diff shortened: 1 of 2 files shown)


types/test.tsx
```diff
@@ -35,13 +36,18 @@ export async function testPureRender() {
 
   // helpers
   const {container, rerender, debug} = page
+  expectType<HTMLElement, typeof container>(container)
```
eps1lon, line 39: This was `Element` before this change which caused https://github.com/testing-library/react-testing-library/pull/834#issuecomment-767796456

nickmccurdy:
Out of curiosity, why not default the return type to `Element` so that SVGs are implicitly supported (even if TypeScript can't infer it properly) and let the user cast it to `HTMLElement` when needed? Could we release this as a temporary fix and then go back after a major release? I think it's valuable to have a more generic return type for accuracy with React (as it can render both HTML and SVG elements).

types/test.tsx
```diff
@@ -35,13 +36,18 @@ export async function testPureRender() {
...
   return {container, rerender, debug}
 }
 
 export function testRenderOptions() {
   const container = document.createElement('div')
   const options = {container}
-  render(<div />, options)
+  const {container: returnedContainer} = render(<div />, options)
+  // Unclear why TypeScript infers `HTMLElement` here when the the input `container` is `HTMLDivElement`
```
nickmccurdy, line 47: This is a known issue with TypeScript. JSX syntax unfortunately can't be inferred to a specific element type. You'd have to use the `React.createElement` syntax, though it's not recommended in most cases.

types/test.tsx
```diff
@@ -35,13 +36,18 @@ export async function testPureRender() {
...
+  // Hopefully this breaks someday and we can switch to
+  // expectType<HTMLDivElement, typeof returnedContainer>(returnedContainer)
+  expectType<HTMLElement, typeof returnedContainer>(returnedContainer)
```
nickmccurdy, line 50: This is a limitation of TypeScript's JSX syntax support. JSX elements unfortunately can't be inferred to a specific element type. You'd have to use the `React.createElement` syntax, though it's not recommended in most cases.

So for now, this test is accurate, though you could avoid JSX if you wanted to still test the regression.

eps1lon: The `<div />` is irrelevant here as far as I know though definitely confusing. The type should be inferred from `options`. 

The confusing part is why it works for `SVGSVGElement` but not `HTMLDivElement`. 

Let me replace `<div />` with `<button />` to check if your assumption is true.

nickmccurdy: Hm, I missed that, I'm curious what happens

eps1lon:
> Out of curiosity, why not continue to default the return type to Element so that SVGs are implicitly supported (even if TypeScript can't infer it properly)

I expanded the test for svg containers to confirm this. They now return `SVGSVGElement` instead of `Element`. So they're still assignable to `Element`.

> Alternatively, could we release this as a temporary fix and then go back after a major release?

Why would we go back?

types/test.tsx
eps1lon: Changed. Is it now clearer that the `container` is not inferred from the rendered UI but the `options`?

types/test.tsx
```diff
@@ -35,13 +36,19 @@ export async function testPureRender() {
...
+  expectType<HTMLElement, typeof container>(container)
   return {container, rerender, debug}
 }
 
 export function testRenderOptions() {
   const container = document.createElement('div')
   const options = {container}
-  render(<div />, options)
+  const {container: returnedContainer} = render(<button />, options)
+  // Unclear why TypeScript infers `HTMLElement` here when the the input `container` is `HTMLDivElement`.
```
timdeschryver, line 47: If we'd swap the [`render` methods in the declaration file](https://github.com/testing-library/react-testing-library/blob/b9836da4d4d2c26494f53b3b6cb629799774615b/types/index.d.ts#L49-L59), this would infer correctly.
I'm unfamiliar with the types for RTL, so I'm not sure if this would break something, or if we could maybe entirely remove the other `render` type.

```suggestion
  // Unclear why TypeScript infers `HTMLElement` here when the input `container` is `HTMLDivElement`.
```

eps1lon: @timdeschryver Awesome, thanks! Fixed in [`a61a0d5` (#868)](https://github.com/testing-library/react-testing-library/pull/868/commits/a61a0d5f7da0acdb72312281d3d02e225e30d288)

timdeschryver: I'm glad I could help 🙂

timdeschryver (approved):

eps1lon:
Merging despite pending codecov/project status. [The codecov comment](https://github.com/testing-library/react-testing-library/pull/868#issuecomment-767824683) is up-to-date and reports the desired 100%.

SimpleCookie:
You know this update was a breaking change right? Not a patch.
My TS-linter is now showing errors, because the types do not match 

`const findByLabelText: (text: Matcher, options?: SelectorMatcherOptions, waitForElementOptions?: waitForOptions) => Promise<HTMLElement>`

> Argument of type 'HTMLElement' is not assignable to parameter of type 'HTMLInputElement | HTMLTextAreaElement'.
  Type 'HTMLElement' is missing the following properties from type 'HTMLTextAreaElement': autocomplete, cols, defaultValue, dirName, and 26 more.ts(2345)

Any idea how to solve it in a nice way?
I am not keen in type casting it like `await findByLabelText("Inskrivningsdag") as HTMLInputElement` everywhere. That's just too tedious.Or perhaps it's a problem with `@testing-library/user-event` instead?

PR #898, opened 2021-04-10
[RFC] fix(fireEvent): Flush effects from discrete events immediately

eps1lon:
I get strong enzyme vibes from this PR (considering we rely on React implementation details). This PR is more exploratory and any discussion should probably be directed at https://github.com/facebook/react/pull/21202 where I'd want this fix to land ideally.

**BREAKING CHANGE**
Importing from `@testing-library/react` will no longer wrap event
dispatches of `@testing-library/dom` in `act(() => {})`.

Essentially reverting https://github.com/testing-library/react-testing-library/pull/685. This will affect `user-event` but I don't see any other way. Wrapping in `act(() => {})` in `user-event` is event worse since `user-event` is explicitly about user events which `fireEevent.click` was not about.

**What**:

If effects in a portaled component were scheduled from discrete event dispatches (e.g. fireEevent.click) then these effects were not flushed during that event but before we returned. This behavior does not match the behavior of events in response to a user interaction.

**Why**:

`fireEvent` majority usage is to test event dispatches in response to user interactions.

**How**:

discrete events are no longer wrapped in `act(() => {})`. Their updates are still flushed when we return though.

**Checklist**:

   I don't think we need to add docs since automatic `configure` was never documented
- [x] Tests
- [ ] Ready to be merged
      PR must be revisited once https://github.com/facebook/react/pull/21202 is resolved.

### DIFF
--- a/src/__tests__/events.js
+++ b/src/__tests__/events.js
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import * as ReactDOM from 'react-dom'
 import {render, fireEvent} from '../'
 
 const eventTypes = [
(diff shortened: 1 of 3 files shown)


eps1lon:
Will be handled in React 18

PR #900, opened 2021-04-21
chore: add tests for Node 16

MichaelDeBoey:
https://nodejs.medium.com/node-js-16-available-now-7f5099a97e70


eps1lon (approved):
Thanks!

Updated branch protection rules to require 16 instead of 15

PR #909, opened 2021-05-11
docs: improve JSDocs for RenderOptions

ImADrafter:
**What**: Improve JS docs for type definitions, including JSDocs for RenderOptions.

*Im just not sure if `docs` commit type is the proper for this one*

**Why**: While using react-testing library, sometimes you will need to review the docs. By filling props description with just a brief stract of the docs, the library would be easier to use for IDEs with JSDoc support, specially for newcommers.

**How**: JSDocs syntax

**Checklist**:

- [ ] Tests
- [x] Ready to be merged


types/index.d.ts
```diff
@@ -34,10 +34,38 @@ export interface RenderOptions<
   Q extends Queries = typeof queries,
   Container extends Element | DocumentFragment = HTMLElement
 > {
+  /** By default, React Testing Library will create a div and append that div to the document.body and this is 
+   *  where your React component will be rendered. If you provide your own HTMLElement container via this option,
+   *  it will not be appended to the document.body automatically.
+   *
+   *  For example: If you are unit testing a tablebody element, it cannot be a child of a div. In this case, you can
```
eps1lon, line 41:
```suggestion
   *  For example: If you are unit testing a `<tbody>` element, it cannot be a child of a div. In this case, you can
```

types/index.d.ts
```diff
@@ -34,10 +34,38 @@ export interface RenderOptions<
   Q extends Queries = typeof queries,
   Container extends Element | DocumentFragment = HTMLElement
 > {
+  /** By default, React Testing Library will create a div and append that div to the document.body and this is 
+   *  where your React component will be rendered. If you provide your own HTMLElement container via this option,
```
eps1lon, line 38:
```suggestion
  /** By default, React Testing Library will create a div and append that div to the document.body. Your React component will be rendered in the created div. If you provide your own HTMLElement container via this option,
```

types/index.d.ts
```diff
@@ -34,10 +34,38 @@ export interface RenderOptions<
...
+   *  it will not be appended to the document.body automatically.
+   *
+   *  For example: If you are unit testing a tablebody element, it cannot be a child of a div. In this case, you can
+   *  specify a table as the render container.
+   * 
+   *  [Container](https://testing-library.com/docs/react-testing-library/api/#container)
+   */
   container?: Container
+  /** If the container is specified, then this defaults to that, otherwise this defaults to document.body. This is used as
```
eps1lon, line 47:
```suggestion
  /** Defaults to the container if the container is specified. Otherwise `document.body` is used for the default. This is used as 
```

types/index.d.ts
```diff
@@ -34,10 +34,38 @@ export interface RenderOptions<
...
+   *  it will not be appended to the document.body automatically.
+   *
+   *  For example: If you are unit testing a tablebody element, it cannot be a child of a div. In this case, you can
+   *  specify a table as the render container.
+   * 
+   *  [Container](https://testing-library.com/docs/react-testing-library/api/#container)
+   */
   container?: Container
+  /** If the container is specified, then this defaults to that, otherwise this defaults to document.body. This is used as 
+   *  the base element for the queries as well as what is printed when you use debug().
```
eps1lon, line 48: formatting:
```suggestion
   *  the base element for the queries as well as what is printed when you use `debug()`.
```

types/index.d.ts
```diff
@@ -34,10 +34,38 @@ export interface RenderOptions<
...
+   *  
+   *  [BaseElement](https://testing-library.com/docs/react-testing-library/api/#baseelement)
+   */
   baseElement?: Element
+  /** If hydrate is set to true, then it will render with ReactDOM.hydrate. This may be useful if you are using server-side
```
eps1lon, line 53:
```suggestion
  /** If `hydrate` is set to `true`, then it will render with `ReactDOM.hydrate`. This may be useful if you are using server-side
```

types/index.d.ts
```diff
@@ -34,10 +34,38 @@ export interface RenderOptions<
...
+   *  For example: If you are unit testing a tablebody element, it cannot be a child of a div. In this case, you can
+   *  specify a table as the render container.
+   * 
+   *  [Container](https://testing-library.com/docs/react-testing-library/api/#container)
+   */
   container?: Container
+  /** If the container is specified, then this defaults to that, otherwise this defaults to document.body. This is used as 
+   *  the base element for the queries as well as what is printed when you use debug().
+   *  
+   *  [BaseElement](https://testing-library.com/docs/react-testing-library/api/#baseelement)
```
eps1lon, line 50: Not a fan of these non-descriptive links. What do you think about just 
```suggestion
   * @see https://testing-library.com/docs/react-testing-library/api/#baseelement
```

eps1lon (changes requested):
Awesome. Really looking forward to having more in-editor docs. I have some minor wording/formatting suggestions.

types/index.d.ts
ImADrafter: Hmm what do you think about this `@links` ? The think with `@see` is that you loose the hyperlink on the hint :( @eps1lon

eps1lon: > The think with @see is that you loose the hyperlink on the hint :(

Will check. I was under the impression VSCode did make these clickable.

3imed-jaberi (approved):
+1

PR #911, opened 2021-05-14
fix: Guard against `process` not being defined

jhnns:
Hi all 👋

**What**:

Fixes

```
ReferenceError: process is not defined
    at eval (webpack-internal:///./node_modules/@testing-library/react/dist/@testing-library/react.esm.js:458:1)
```

when @testing-library/react is being used with webpack 5 and karma.

**Why**:

Webpack 5 doesn't shim Node's `process` object by default anymore. Only instances of `process.env.NODE_ENV` are replaced statically. This means that unguarded checks of `process.env` will crash in browser environments.

**How**:

I was wondering if there should also be a check in `dont-cleanup-after-each.js`. I decided against it since this would need to define a global `process` object in order to define the env variable. After all, it's the app developer's responsibility to provide a global `process` object in this case.

**Checklist**:

- [ ] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,7 +5,7 @@ import {cleanup} from './pure'
 // this ensures that tests run in isolation from each other
 // if you don't like this then either import the `pure` module
 // or set the RTL_SKIP_AUTO_CLEANUP env variable to 'true'.
-if (!process.env.RTL_SKIP_AUTO_CLEANUP) {
+if (typeof process === "undefined" || !process.env?.RTL_SKIP_AUTO_CLEANUP) {
   // ignore teardown() in code coverage because Jest does not support it
   /* istanbul ignore else */
   if (typeof afterEach === 'function') {


src/index.js
```diff
@@ -5,7 +5,7 @@ import {cleanup} from './pure'
 // this ensures that tests run in isolation from each other
 // if you don't like this then either import the `pure` module
 // or set the RTL_SKIP_AUTO_CLEANUP env variable to 'true'.
-if (!process.env.RTL_SKIP_AUTO_CLEANUP) {
+if (typeof process === "undefined" || !process.env?.RTL_SKIP_AUTO_CLEANUP) {
```
jhnns, line 8: Does your build setup support optional chaining? I hope that's fine 😁

eps1lon (approved):
Thanks!

src/index.js
eps1lon: Should be transpiled loosely.

eps1lon:
@all-contributors Add @jhnns for code

PR #922, opened 2021-06-08
chore: Don't test with node 15

eps1lon:
**What**:

Don't test with node 15

**Why**:

Node 15 reached EOL on June, 1st so let's free up some CI ressources.

**How**:

Remove 15 from node test matrix.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -17,7 +17,7 @@ jobs:
     if: ${{ !contains(github.head_ref, 'all-contributors') }}
     strategy:
       matrix:
-        node: [10.13, 12, 14, 15, 16]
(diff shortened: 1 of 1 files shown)


.github/workflows/validate.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     if: ${{ !contains(github.head_ref, 'all-contributors') }}
     strategy:
       matrix:
-        node: [10.13, 12, 14, 15, 16]
+        node: [10.13, 12, 14, 16]
```
nickmccurdy, line 20: We might as well drop 10 as well since support ended on April 30, right?

Issue #509, opened 2019-10-25
Support for React 18


FredyC:
### Describe the feature you'd like:

It's probably premature, given [experimental nature of Concurrent mode](https://reactjs.org/docs/concurrent-mode-intro.html). but I might as we start the discussion.

Perhaps I am misunderstanding something, but I think it would be useful if tests would run in Concurrent mode to mimic real app behavior as close as possible. 

At this point, it's probably about replacing `ReactDOM.render` with `ReactDOM.createRoot`. 

The question is how to make it in a compatible way so people can decide which version to use without polluting actual test files with it.

### Suggested implementation:

Probably the most obvious choice is exposing function `renderConcurrent` which would be fine for the experimental phase. However, given that Concurrent will most likely become a "new normal" eventually, it would be kinda weird to have that in tests.

Perhaps the env variable might be better approach ultimately. It's easy to switch and doesn't need to pollute actual test files with it.

### Describe alternatives you've considered:

Publishing major version when Concurrent mode is official and stable and keep rolling in that direction only. An obvious drawback is that people would need to stick to the previous major as long as they are testing against code that doesn't support Concurrent mode.

eps1lon:
This is how https://github.com/testing-library/react-testing-library/pull/507 looks so far:

```jsx
 const {container} = render(<div>test</div>, {root: 'concurrent'})

  expect(container).toHaveTextContent('test')
```

There's no breaking change required so far. It's more about figuring out how to teach this (do we need to test differently? what scheduler implementations are useful for testing? etc).

FredyC:
Yea well, that's another "obvious choice", but I don't think it's a correct approach. Sure it gives good choice that only some tests can run concurrently. However, it's too verbose in my opinion.

I am still convinced that using the environment variable is the best approach here. It's then possible to pick which test files will run concurrently and which not by having two sets of `test` script.

eps1lon:
> Yea well, that's another "obvious choice", but I don't think it's a correct approach. Sure it gives good choice that only some tests can run concurrently. However, it's too verbose in my opinion.

I mean we can always read the default from a environment variable (or you import your own renderer that does that). 

Your approach would put the restriction on you to have different test files for different react modes. Mine would have no such restriction. This becomes more obvious if you think about blocking vs concurrent roots. You might want to test them next to each other.

I don't understand how this is different to, say, `hydrate`. That would be too verbose as well in your opinion?

FredyC:
[...]

PR #925, opened 2021-06-08
feat: Use concurrent React when available

eps1lon:
TODO:
- [x] how to tell the coverage report that we've reached full coverage across the whole build matrix i.e. with different versions of React?
   By telling codecov that different React versions act as [feature flags](https://docs.codecov.io/docs/merging-reports)

Revisiting https://github.com/testing-library/react-testing-library/pull/507

BREAKING CHANGE: Concurrent features will be used by default if supported by the installed version of `react-dom`. To opt-out use `render(element, { legacyRoot: true }).

**What**:

Closes https://github.com/testing-library/react-testing-library/issues/509

**Why**:

Concurrent features will be available in React 18.

Opt-out (breaking change) vs opt-in (feature release):
`createRoot` is considered to be the default in React 18 so we should follow.

**How**:

Add a `concurrent` option to `render`. Note that there are still some unsolved problems with testing concurrent features like https://github.com/reactwg/react-18/discussions/21#discussioncomment-796755 and https://github.com/reactwg/react-18/discussions/23#discussioncomment-798801. So testing concurrent features will change. We will only support the latest pre-release of React 18. So any release of `@testing-library/react` may break your testing setup if don't have the latest alpha/beta/release candidate installed.

**Checklist**:

- [x] Tests
- [ ] Ready to be merged


kentcdodds:
> how to tell the coverage report that we've reached full coverage across the whole build matrix i.e. with different versions of React?

I believe that codecov can combine coverage reports, but I don't think there's a reasonable way to make Jest happy about our coverage number in a single environment 😬

src/pure.js
```diff
@@ -99,25 +130,91 @@ function render(
...
+    // default to document.body instead of documentElement to avoid output of potentially-large
+    // head elements (such as JSS style blocks) in debug output
+    baseElement = document.body
+  }
+  if (!container) {
+    container = baseElement.appendChild(document.createElement('div'))
+  }
+
+  let root
+  // eslint-disable-next-line no-negated-condition -- we want to map the evolution of this over time. The root is created first. Only later is it re-used so we don't want to read the case that happens later first.
```
kentcdodds, line 154: I think the rule is only complaining about the fact that this code could be written like this:

```tsx
if (mountedContainers.has(container)) {
  // the forEach here
} else {
  // create the root impl here.
}
```

src/pure.js
```diff
@@ -99,25 +130,91 @@ function render(
...
+}
+
+function waitForElementToBeRemoved(callback, options) {
+  return waitForElementToBeRemovedDTL(() => {
+    let result
+    act(() => {
+      result = callback()
+    })
+    return result
+  }, options)
```
kentcdodds, line 212: I take it that these are added here because the removal of `asyncWrapper`?

src/pure.js
```diff
@@ -4,63 +4,90 @@ import {
...
 import {fireEvent} from './fire-event'
 
 configureDTL({
-  asyncWrapper: async cb => {
-    let result
-    await asyncAct(async () => {
-      result = await cb()
-    })
-    return result
-  },
```
kentcdodds, line 18: So to be clear, there's no longer the concept of an `async` act anymore?

tests/setup-env.js
```diff
@@ -1 +1,3 @@
 import '@testing-library/jest-dom/extend-expect'
+// eslint-disable-next-line import/no-extraneous-dependencies -- need the version from React not an explicitly declared one
+jest.mock('scheduler', () => require('scheduler/unstable_mock'))
```
kentcdodds, line 3: I thought mocking the scheduler wasn't required: https://github.com/reactwg/react-18/discussions/23#discussioncomment-798952

kentcdodds (commented):
Thanks for your work on this and for your work in the working group! I have a few questions.

nickmccurdy:
I've just read about [the plans for React 18.0.0 alpha](https://reactjs.org/blog/2021/06/08/the-plan-for-react-18.html), and I was thinking about testing out the new `createRoot` renderer. Do you think it would be better to do that in a different PR or get this PR working with both the legacy and concurrent versions of React 18 alpha? As far as I understand the legacy version of React 18 should behave very similarly to React 17, though if we still needed to test both we could adjust our build matrix.

nickmccurdy:
Since React 18 is publishing new releases nightly to both the next and experimental tags, should we only use one of those two tags for now (to save CI resources as they're both the same builds of React 18 alpha) and run daily scheduled builds (so we can catch React alphas causing a breakage even if there isn't a commit to Testing Library)? I can try this in a separate PR if you'd like.

src/pure.js
eps1lon: This was my interpretation as well. But I wanted to map the code to the lifetime of the root/container tuple as well: 
1. gets created
2. gets reused

ESLint is telling me to put the "gets reused" branch first which is not what happens in time. So if you're reading the code top-down you don't know where the root/container tuple is coming from. 

I should rewrite the comment since I tried to explain the reason there. 

Basically, I don't think `no-negated-condition` does anything useful. I could just as well write

```js
const needsRoot = !mountedContainers.has(container)
if (needsRoot) {
  // create the root impl here.
} else {
  // the forEach here
}
```
which means I have to name one more thing but then I don't have to use the noisy eslint-disable with confusing comments.

src/pure.js
eps1lon: Yeah, updates aren't flushed until we exit the outermost act. `asyncWrapper` was wrapping the whole `waitFor` call so the screen didn't ever get updated. 

I'm not sure if we shouldn't actually change the dom-testing-library implementation so that we end up with 

```js
act(() => {
  jest.advanceTimersByTime(interval)
})
```

or fork it entirely. 

And we'd still have the issue of real timers: https://github.com/reactwg/react-18/discussions/23#discussioncomment-812450

src/pure.js
eps1lon: > So to be clear, there's no longer the concept of an async act anymore?

Nothing changed in that regard. 

I needed to remove to wrapper because  it was used to wrap all the polling in `waitFor`:

```js
await ReactDOMTestUtils.act(async () => {
  await new Promise((resolve) => {
    setInterval(async () => {
      if (
        // That's the condition we wait for
        document.querySelector("main").getAttribute("aria-busy") !== "true"
      ) {
        resolve();
      }
    }, 50);
  });
});
```

and this meant that the screen isn't actually updated during the polling i.e. it'll always reject and then the screen gets updated. This may be fine since we could just check one more time after the timeout but then `interval` becomes useless and the function runtime is always equal to `timeout`.

tests/setup-env.js
eps1lon: It was my understanding that these changes are planned but didn't land yet. At least it was required a week ago and I didn't see anything change. Let me check again and then add a TODO linking the discussion comment.

.github/workflows/validate.yml
```diff
@@ -16,6 +16,7 @@ jobs:
     # ignore all-contributors PRs
     if: ${{ !contains(github.head_ref, 'all-contributors') }}
     strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
eps1lon: I would land this change regardless of this PR. Personally, I'd like to know if a job only fails in one job. Otherwise how do I know what's causing the issue?

src/pure.js
```diff
@@ -4,63 +4,90 @@ import {
...
   eventWrapper: cb => {
     let result
     act(() => {
-      result = cb()
+      // TODO: Remove ReactDOM.flushSync once `act` flushes the microtask queue.
+      // Otherwise `act` wrapping updates that schedule microtask would need to be followed with `await null` to flush the microtask queue manually
+      // See https://github.com/reactwg/react-18/discussions/21#discussioncomment-796755
+      result = ReactDOM.flushSync(cb)
```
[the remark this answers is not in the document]
eps1lon: From my understand of https://github.com/reactwg/react-18/discussions/21#discussioncomment-798691 and the thread it spawned (which is linked here) these are planned changes so we still need to flush sync.

src/pure.js
```diff
@@ -99,25 +130,91 @@ function render(
...
 }
 
-function cleanup() {
-  mountedContainers.forEach(cleanupAtContainer)
+function render(
+  ui,
+  {
+    container,
+    baseElement = container,
+    legacyRoot = typeof ReactDOM.createRoot !== 'function',
```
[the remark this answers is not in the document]
eps1lon: > Defaulting to using createRoot could cause unexpected test failures for users that upgrade from React 17 to 18

I think this risk will be eventually very low. In the end this PR will land in a major release so breakage should be expected. I just don't think it's healthy if React considers one API the default while we consider another one the default. Ultimately you can always wrap our render with a different default value. In `jest` you could just mock it which is what I did originally: https://github.com/eps1lon/react-testing-library/blob/04e865c0009a0d709a76846c80cdda683ac4b0fc/src/__mocks__/pure.js

eps1lon:
> I believe that codecov can combine coverage reports, but I don't think there's a reasonable way to make Jest happy about our coverage number in a single environment

This seems sufficient to me.

eps1lon:
> Since React 18 is publishing new releases nightly to both the next and experimental tags, should we only use one of those two tags for now (to save CI resources as they're both the same builds of React 18 alpha) and run daily scheduled builds (so we can catch regressions in React alphas even if there isn't a commit to Testing Library)?

As far as I know experimental and next release channels will stay so I wouldn't want to change something  temporarily. Especially since I'm not aware of any CI ressource problems we have.

tests/setup-env.js
eps1lon: Still warns without the mocked scheduler (message is even outdated):

> Warning: In Concurrent or Sync modes, the "scheduler" module needs to be mocked to guarantee consistent behaviour across tests and browsers. For example, with jest: 
      jest.mock('scheduler', () => require('scheduler/unstable_mock'));
      

Added a TODO comment to make this clear.

src/pure.js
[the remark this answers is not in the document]
eps1lon: Confirmed `flushSync` is still needed. Otherwise most of the event related tests fail.

eps1lon:
> > I believe that codecov can combine coverage reports, but I don't think there's a reasonable way to make Jest happy about our coverage number in a single environment
> 
> This seems sufficient to me. Right now I'm just guessing that we actually cover everything across next and experimental.

Simpler than expected: [`3aa9ed9` (#925)](https://github.com/testing-library/react-testing-library/pull/925/commits/3aa9ed93a90267eeedb96e0d832cb3fb3be4bae9)

nickmccurdy:
> > Since React 18 is publishing new releases nightly to both the next and experimental tags, should we only use one of those two tags for now (to save CI resources as they're both the same builds of React 18 alpha) and run daily scheduled builds (so we can catch regressions in React alphas even if there isn't a commit to Testing Library)?
> 
> As far as I know experimental and next release channels will stay so I wouldn't want to change something temporarily. Especially since I'm not aware of any CI ressource problems we have.

What do you think about using [scheduled events] for daily builds (without changing the tags)?

kentcdodds:
FYI, I'm going out of town for a week so don't hold this up on my account.

src/pure.js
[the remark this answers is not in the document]
eps1lon: Small update: React now logs a deprecation warning in React 18 if you use `ReactDOM.render` or `ReactDOM.hydrate`. I think we should not trigger warnings by default so the "`createRoot` by default" approach is going in the right direciton in my opinion.

eps1lon:
> FYI, I'm going out of town for a week so don't hold this up on my account.

The reason I'm hesitant is that `waitFor` and `findBy` doesn't really work anymore with React 18. I'm hoping this gets answered soon but I'd rather release the current state of `alpha` first and defer this PR to another major release.

eps1lon:
No github the branch is there again so why can't I re-open this? Well thanks for loosing context. Continued in https://github.com/testing-library/react-testing-library/pull/937

PR #937, opened 2021-06-23
feat: Use concurrent React when available

eps1lon:
BREAKING CHANGE: If you have React 18 installed, we'll use the new [`createRoot` API](https://github.com/reactwg/react-18/discussions/5) by default which comes with a set of [changes while also enabling support for concurrent features](https://github.com/reactwg/react-18/discussions/4).
To can opt-out of this change by using `render(ui, { legacyRoot: true } )`. But be aware that the legacy root API is deprecated in React 18 and its usage will trigger console warnings.

If you want to try out this build follow the "local install instructions" in https://ci.codesandbox.io/status/testing-library/react-testing-library/pr/937/

Continuation of https://github.com/testing-library/react-testing-library/pull/925 that GitHub is unable to re-open.

Integration tests:
- https://github.com/mui-org/material-ui/pull/28267

**What**:

Closes https://github.com/testing-library/react-testing-library/issues/509

**Why**:

Concurrent features will be available in React 18.

Opt-out (breaking change) vs opt-in (feature release):
`createRoot` is considered to be the default in React 18 so we should follow.

**How**:

Add a `concurrent` option to `render`. Note that there are still some unsolved problems with testing concurrent features like https://github.com/reactwg/react-18/discussions/21#discussioncomment-796755 and https://github.com/reactwg/react-18/discussions/23#discussioncomment-798801. So testing concurrent features will change. We will only support the latest pre-release of React 18. So any release of `@testing-library/react` may break your testing setup if don't have the latest alpha/beta/release candidate installed.

**Checklist**:

- [x] Tests
- [x] Ready to be merged


src/pure.js
```diff
@@ -25,42 +18,83 @@ configureDTL({
   },
 })
 
+if (typeof React.startTransition !== undefined) {
+  configureDTL({
+    unstable_advanceTimersWrapper: cb => {
+      return act(cb)
+    },
+    asyncWrapper: cb => cb(),
```
kentcdodds, line 26: Is this necessary to configure?

kentcdodds (approved):
Looks solid! Just one question.

src/pure.js
eps1lon: TL;DR: I misunderstood the question. I think we don't need to configure that because it's the default. I still leave my original long winded explainer for future reference and because it made me realize I still don't fully understand why `act()` behaves the way it does.

So let's say we have 

```js
waitFor(() => expect(getByRole('button')).toBeInTheDocument())
```
and the button would only appear after two intervals. This would previously translate to 
```js
act(() => {
   jest.advanceTimersByTime();
   checkCallback();
   jest.advanceTimersByTime();
   checkCallback();
});
```
The problem is that React only flushes effects when exiting the `act` call. So `checkCallback` could not assert on any effect scheduled due to timers being advanced.
If we just configure `unstable_advanceTimersWrapper` to use `act` we would now have
```
act(() => {
   jest.advanceTimersByTime();
   act(() => {
     checkCallback();
   })
   jest.advanceTimersByTime();
   act(() => {
     checkCallback();
   });
});
```
The problem now is that React only flushes scheduled effects in the outermost act so we need to remove that.

I think I good question is why only the outermost `act` is considered.

eps1lon: I need to double check something though. React 17 will no longer have `act` as the `asyncWrapper` but somehow our React 17 tests are still passing.

MatanBobi: > I need to double check something though. React 17 will no longer have `act` as the `asyncWrapper` but somehow our React 17 tests are still passing. So either our tests are incomplete or this was dead code.

Did you mean React 18? If so, following [this comment](https://github.com/reactwg/react-18/discussions/23#discussioncomment-1087260) I'm not sure it won't work, it's just not the recommendation, isn't it?

eps1lon: > Did you mean React 18? If so, following this comment I'm not sure it won't work, it's just not needed anymore, isn't it?

No, 17. I was reffering to our usage though. I changed the behavior for React 17 as well but no tests started failing.

nickmccurdy (approved):

src/pure.js
eps1lon: > I need to double check something though. React 17 will no longer have `act` as the `asyncWrapper` but somehow our React 17 tests are still passing. So either our tests are incomplete or this was dead code.

Tests are incomplete due to usage of class components. Backported the new tests with a function component in https://github.com/testing-library/react-testing-library/pull/962 that should ensure `asyncWrapper` is still present in React 17.

eps1lon: Prevented mild disaster since the new tests also uncovered silly lines such as `typeof React.startTransition !== undefined`. It's either `React.startTransition !== undefined` or `typeof React.startTransition !== 'undefined'`.

Also uncovered that the "configure wrapper" approach is not compatible with supporting different timers across multiple versions of react.

eps1lon: Going to leave this a bit until the next WG meeting. Unclear what the timeline is for work on act. Feels like the core team already moved on but there are still glaring holes in the behavior of act().

eps1lon: Though I'm personally fine with shipping this as an alpha to test integration. We haven't solved all problems yet for the alpha but neither has React itself. I'm fine with shipping this in an incomplete state since I can just blame React about all the missing parts 😎

kentcdodds: Sounds fine to me 👍

nstepien:
Testing this `13.0.0-alpha.1` release in https://github.com/adazzle/react-data-grid/pull/2459 and now React is printing a ton of `act` warnings, is that expected?
https://github.com/adazzle/react-data-grid/runs/3585365674

diegohaz:
> Testing this `13.0.0-alpha.1` release in [adazzle/react-data-grid#2459](https://github.com/adazzle/react-data-grid/pull/2459) and now React is printing a ton of `act` warnings, is that expected?
> https://github.com/adazzle/react-data-grid/runs/3585365674

@nstepien I'm also experiencing this on tests using `waitFor`. But, if I understood correctly, this hasn't been fully implemented in this alpha release because of https://github.com/testing-library/react-testing-library/pull/937#discussion_r706618576

eps1lon:
@nstepien @diegohaz Please read https://github.com/testing-library/react-testing-library/issues/509#issuecomment-917989936

PR #945, opened 2021-07-24
Add mocha-watch-cleanup-after-each Mocha hook

jankalfus:
**What**: Added a hook which can be used in Mocha's watch mode to do automated cleanup in tests, based on discussion in https://github.com/testing-library/testing-library-docs/pull/887

**Why**: Without manual cleanup, or a custom root hook, running tests written using react-testing-library using Mocha results in test fails, as no cleanup is done by react-testing-library on subsequent runs, so elements from previous runs remain in the DOM.

**How**: Added a Mocha root hook which runs cleanup. The hook can be used when running `mocha --watch` (or `mocha` without watch).

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs): https://github.com/testing-library/testing-library-docs/pull/887
- [ ] Tests - I don't believe I can add any automated test without also adding mocha, which I don't think is desired. Anyway, I've tested the change locally in a project I have mocha set up in and it works great.
- [ ] TypeScript definitions updated - N/A
- [X] Ready to be merged

### DIFF
--- a/mocha-cleanup-after-each.js
+++ b/mocha-cleanup-after-each.js
@@ -1,0 +1,8 @@
+require('./dont-cleanup-after-each')
+const rtl = require('./')
+
+exports.mochaHooks = {
+  afterEach() {
+    rtl.cleanup()
+  },
+}
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
   "files": [
     "dist",
     "dont-cleanup-after-each.js",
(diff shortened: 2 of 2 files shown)


eps1lon:
This seems small enough to let users do themselves. We already support jest and we don't have the bandwidth to extend that support in this library to yet another runner.

A documentation only change is sufficient.

PR #966, opened 2021-09-24
fix(TS): make wrapper allow a simple function comp

### LINKED ISSUE #965: Bug: Render Function Wrapper Typescript Issue
- `@testing-library/react` version: ^11.1.0
- Testing Framework and version(JEST): ^26.6.0 
- Dom ENV: JS dom(default, no specific config)

### Relevant code or config:

```js
import { render as rtlRender, RenderOptions, screen } from '@testing-library/react';
import { createMemoryHistory } from 'history';
import React, { Component, ReactElement } from 'react';
import { Router } from 'react-router-dom';
import App from './App';

type DefaultOptions = Omit<RenderOptions, 'queries'> & { route: string };

const render = (ui: ReactElement, options?: DefaultOptions) => {
  const history = createMemoryHistory({ initialEntries: [options?.route || '/'] });

  const Wrapper = ({ children }: { children: ReactElement }) => {
    return <Router history={history}>{children}</Router>;
  };

  return rtlRender(ui, { wrapper: Wrapper, ...options });
};

test('renders to screen', () => {
  render(<App />);

  expect(screen.getByTestId('router')).toBeInTheDocument();
});
```

### What happened:
This is a typescript error. check the screenshot below for the error:
Screenshot 2021-09-24 at 12 35 12

### Problem description:

I believe the problem is coming from here:
Screenshot 2021-09-24 at 12 36 41

### Suggested solution:
I believe that this can be solved easily by a simple change. I've changed the wrapper type from `React.ComponentType` to `JSX.Element`. Is this a bug or is this being done intentionally?

Screenshot 2021-09-24 at 12 37 32

kentcdodds:
**What**: This adds an option for the wrapper to use a simple function component

**Why**: Closes #965

**How**: Added the option to the types for `render`

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -70,7 +70,7 @@ export interface RenderOptions<
    *
    *  @see https://testing-library.com/docs/react-testing-library/api/#wrapper
    */
-  wrapper?: React.ComponentType | (props: {children: React.ReactNode}) => JSX.Element
+  wrapper?: React.ComponentType | ((props: {children: React.ReactNode}) => JSX.Element)
 }
 
 type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -6,7 +6,6 @@ import {
   BoundFunction,
   prettyFormat,
 } from '@testing-library/dom'
-import React from 'react'
 import {Renderer} from 'react-dom'
 import {act as reactAct} from 'react-dom/test-utils'
 
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -6,6 +6,7 @@ import {
   BoundFunction,
   prettyFormat,
 } from '@testing-library/dom'
+import React from 'react'
 import {Renderer} from 'react-dom'
 import {act as reactAct} from 'react-dom/test-utils'
 
@@ -70,9 +71,7 @@ export interface RenderOptions<
    *
    *  @see https://testing-library.com/docs/react-testing-library/api/#wrapper
    */
-  wrapper?:
-    | React.ComponentType
-    | ((props: {children: React.ReactNode}) => JSX.Element)
+  wrapper?: React.ComponentType<{children: React.ReactElement}>
 }
 
 type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -70,7 +70,7 @@ export interface RenderOptions<
    *
    *  @see https://testing-library.com/docs/react-testing-library/api/#wrapper
    */
-  wrapper?: React.ComponentType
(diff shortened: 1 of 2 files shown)


kentcdodds:
I don't like merging my own PRs, especially in projects I'm not actively maintaining 😅 So I can wait until someone else can get around to this one. Thanks!

3imed-jaberi (dismissed):
+1 🚀

types/index.d.ts
```diff
@@ -70,7 +70,9 @@ export interface RenderOptions<
    *
    *  @see https://testing-library.com/docs/react-testing-library/api/#wrapper
    */
-  wrapper?: React.ComponentType
+  wrapper?:
+    | React.ComponentType
+    | ((props: {children: React.ReactNode}) => JSX.Element)
```
eps1lon, line 75: Let me add a type test for this because the fix is simpler I believe.

eps1lon (changes requested):

types/index.d.ts
eps1lon: Implicit children strikes again:

`React.ComponentType` includes `React.FunctionComponent` which means the accepted props are `{ children?: React.ReactNode }`. However, our `Wrapper` only accepts `{ children: React.ReactElement }` props. So we could render `<Wrapper>Some Child</Wrapper>` which may not be implemented by the custom `Wrapper` component because now it receives `{ children: string }`.

The proposed fix would not have fixed the linked issue (`npm typecheck` fails on ffa222170ab0f4654319e611b5f48fce8590ddbf). dc03098550817b9e222473c97fdc40773847c05f fixes the linked issue.

eps1lon (approved):

kentcdodds:
LGTM 👍 Thanks!

astorije:
Hi @eps1lon!

We currently have code that, once contrived a lot for the sake of sharing an example, looks somewhat like this:

```tsx
import { PropsWithChildren } from 'react';

function Foo({ children }: PropsWithChildren<Bar>) {
  return (
    options?.wrapper ? (
      <options.wrapper>{children}</options.wrapper>
    ) : (
      children
    )
  );
}
```

`PropsWithChildren` comes from `@types/react` and is simply:
```ts
type PropsWithChildren<P> = P & { children?: ReactNode | undefined };
```

This worked fine with `@testing-library/react` v12.1.0, but starting in v12.1.2 we're seeing the following, after the type change in this PR:

```
error TS2322: Type 'ReactNode' is not assignable to type 'ReactElement<any, string | JSXElementConstructor<any>>'.
  Type 'undefined' is not assignable to type 'ReactElement<any, string | JSXElementConstructor<any>>'.

             <options.wrapper>{children}</options.wrapper>
```

What would be your recommendation here?

Should we simply change the code above into `<options.wrapper><>{children}</></options.wrapper>`?

(It seems this discussion with @drzamich is also relevant: https://github.com/testing-library/react-testing-library/pull/973/files#r754242764)

Thank you! ❤️

Issue #972, opened 2021-10-02
render wrapper error TS2769: No overload matches this call


akornatskyy:
- `@testing-library/react` version: 12.1.1
- Testing Framework and version: jest@27.2.4
  
- typescript@4.4.3

### Relevant code or config:

```js
import {MemoryRouter as Router} from 'react-router-dom';
// ...
render(<Whatever />, {wrapper: Router});
```

### What you did:

Use `Router` as `wrapper`.

### What happened:

typescript compile error.

### Problem description:

```typescript
 Overload 1 of 2, '(ui: ReactElement<any, string | JSXElementConstructor<any>>, options: RenderOptions<typeof import(".../node_modules/@testing-library/dom/types/queries"), HTMLElement>): RenderResult<...>', gave the following error.
Type 'typeof MemoryRouter' is not assignable to type 'ComponentType<{ children: ReactElement<any, string | JSXElementConstructor<any>>; }> | undefined'.
Type 'typeof MemoryRouter' is not assignable to type 'ComponentClass<{ children: ReactElement<any, string | JSXElementConstructor<any>>; }, any>'.
Types of parameters 'props' and 'props' are incompatible.
Type '{ children: ReactElement<any, string | JSXElementConstructor<any>>; }' is not assignable to type 'MemoryRouterProps | Readonly<MemoryRouterProps>'.
Type '{ children: ReactElement<any, string | JSXElementConstructor<any>>; }' has no properties in common with type 'Readonly<MemoryRouterProps>'.
Overload 2 of 2, '(ui: ReactElement<any, string | JSXElementConstructor<any>>, options?: Omit<RenderOptions<typeof import(".../node_modules/@testing-library/dom/types/queries"), HTMLElement>, "queries"> | undefined): RenderResult<...>', gave the following error.
Type 'typeof MemoryRouter' is not assignable to type 'ComponentType<{ children: ReactElement<any, string | JSXElementConstructor<any>>; }> | undefined'.
Type 'typeof MemoryRouter' is not assignable to type 'ComponentClass<{ children: ReactElement<any, string | JSXElementConstructor<any>>; }, any>'.
```

akornatskyy:
That is still the issue with **12.1.2**.

akornatskyy:
@eps1lon : just in case you have missed the last comment above, this error still can be reproduced with `12.1.2` (I cannot reopen this issue). advise, please.

tb-viktor:
Have you managed to make some progress on this issue?

[...]

Issue #970, opened 2021-09-27
Wrapper type breaks types with props other than children


cfbender:
- `@testing-library/react` version: 12.1.1
- Testing Framework and version: jest @ 27.2.2
  
- DOM Environment: 5.14.1
  

### Relevant code or config:

```ts
type AppWrapperProps = {
  userProviderProps?: UserProviderProps
}

export const AppWrapper: React.FC<AppWrapperProps> = ({
  children,
  userProviderProps = { user: mockUser },
}) => (
  <UserProvider {...userProviderProps}>
...
```

### What you did:

Upgraded @testing-library/react to 12.1.1

### What happened:

Type for wrapper broken in #966

### Reproduction:

Use wrapper component that has some props that can be passed

### Problem description:

The type being inflexible (only allowing `children`) means that the wrapper can't take props if used anywhere else

### Suggested solution:

Something like `wrapper?: React.ComponentType<React.PropsWithChildren>`

eps1lon:
Could you provide a full repro how `AppWrapper` receives all the props with your custom `render` method?

cfbender:
Sure thing!

we are wanting to sometimes use it as such:

```ts
type AppWrapperProps = {
  userProviderProps?: UserProviderProps
}

export const AppWrapper: React.FC<AppWrapperProps> = ({
  children,
  userProviderProps = { user: mockUser },
}) => (
  <UserProvider {...userProviderProps}>
    {children}
  </UserProvider>
)

...

    render(<Page {...defaultPageProps.props} />, {
      wrapper: AppWrapper,
    })
    
    // OR
    
    render(
      <AppWrapper userProviderProps={{ ...user }}>
        <Page {...defaultPageProps.props} />
      </AppWrapper>
    )
```

this gives us the ability to use the same component in multiple ways and overriding some props in the provider

bfaulk96:
Came here with a very similar issue. The latest release caused new TS type mismatch errors

sabertazimi:
Similar issue. Failed CI [case](https://github.com/sabertazimi/blog/actions/runs/1296051909):

```bash
Error: src/components/ErrorBoundary/ErrorBoundary.test.tsx(51,7): error TS2769: No overload matches this call.
[...]

PR #973, opened 2021-10-03
fix(render): Don't reject wrapper types based on statics

eps1lon:
**What**:

Closes https://github.com/testing-library/react-testing-library/issues/970
Closes https://github.com/testing-library/react-testing-library/issues/972

**Why**:

Broke in https://github.com/testing-library/react-testing-library/pull/966

**How**:

Implicit children strikes yet again. While `FunctionComponent` adds implicit children, its `propTypes` do not so we didn't catch that issue. We don't care about React statics and just that we can pass it to `createElement` (or any JSX runtime). 

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -70,7 +70,7 @@ export interface RenderOptions<
    *
    *  @see https://testing-library.com/docs/react-testing-library/api/#wrapper
    */
-  wrapper?: React.ComponentType<{children: React.ReactElement}>
+  wrapper?: React.JSXElementConstructor<{children: React.ReactElement}>
 }
 
 type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
--- a/types/test.tsx
+++ b/types/test.tsx
@@ -115,7 +115,9 @@ export function wrappedRenderB(
   ui: React.ReactElement,
   options?: pure.RenderOptions,
 ) {
-  const Wrapper: React.FunctionComponent = ({children}) => {
+  const Wrapper: React.FunctionComponent<{children?: React.ReactNode}> = ({
+    children,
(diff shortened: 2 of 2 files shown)


types/test.tsx
```diff
@@ -115,13 +115,32 @@ export function wrappedRenderB(
   ui: React.ReactElement,
   options?: pure.RenderOptions,
 ) {
-  const Wrapper: React.FunctionComponent = ({children}) => {
+  const Wrapper: React.FunctionComponent<{children?: React.ReactNode}> = ({
```
eps1lon, line 118: While `FunctionComponent` adds implicit children, its `propTypes` do not so we didn't catch that issue. Here we make sure all statics assume the same props. Also makes sure this doesn't break once we get rid of implicit children in `React.FunctionComponent`.

MatanBobi (approved):
Looks good :)

types/index.d.ts
```diff
@@ -70,7 +70,7 @@ export interface RenderOptions<
    *
    *  @see https://testing-library.com/docs/react-testing-library/api/#wrapper
    */
-  wrapper?: React.ComponentType<{children: React.ReactElement}>
+  wrapper?: React.JSXElementConstructor<{children: React.ReactElement}>
```
drzamich, line 73: I'm wondering whether we should define `children` as `React.ReactNode` which seems to be the "official" type for children props. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L829

drzamich: That makes perfect sense - `ReactNode` is indeed very wide. Thanks for the explanation.

PR #994, opened 2021-11-18
feat: Enable "missing act" warnings in React 18 by default

eps1lon:
**What**:

Enables "missing act" warnings in React 18 for test runners that implement `beforeAll` and `afterAll` (e.g. Jest).
This can be disabled the same way auto-cleanup can be disabled.

**Why**:

Ensures same behavior regarding missing act warnings across as React 16 and 17

**How**:

Set [`IS_REACT_ACT_ENVIRONMENT`](https://github.com/reactwg/react-18/discussions/102) to true once before tests start.

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs) Will wait until React docs for IS_REACT_ACT_ENVIRONMENT are available.
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/act-compat.js
+++ b/src/act-compat.js
@@ -32,7 +32,7 @@ function getGlobalThis() {
   throw new Error('unable to locate global object')
 }
 
-function setReactActEnvironment(isReactActEnvironment) {
+function setIsReactActEnvironment(isReactActEnvironment) {
   getGlobalThis().IS_REACT_ACT_ENVIRONMENT = isReactActEnvironment
 }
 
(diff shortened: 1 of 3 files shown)


eps1lon:
The warning is a pretty important part of the migration story to understand why certain tests might start failing or at least excibit different behavior. At least with React 18 users are able to disable the "missing act" warning if they truly want to so this change should be fairly safe.

The only issue might if people reset their modules in `beforeEach` so we may want to switch to `beforeEach` instead of `beforeAll`.

IanVS:
For anyone else finding this and wondering what the `same way auto-cleanup can be disabled` is, the docs are at: https://testing-library.com/docs/react-testing-library/setup#skipping-auto-cleanup

PR #1001, opened 2021-12-03
fix: Add `@types/react-dom` as a direct dependency

### LINKED ISSUE #1000: `@types/react-dom` incorrectly declared as a `devDependency`
- `@testing-library/react` version: v12.1.2
- Testing Framework and version: N/A
- DOM Environment: N/A

### What you did:

Tried to use `@testing-library/react` in a TS web app built with Bazel's [rules_nodejs](https://github.com/bazelbuild/rules_nodejs), which enforce strict visibility between NPM packages as determined by their declared dependencies.

### What happened:

Got the following tsc error when compiling a test that uses `@testing-library/react` with `rules_nodejs`:
```
node_modules/@testing-library/react/types/index.d.ts(9,24): error
TS2307: Cannot find module 'react-dom' or its corresponding type
declarations.
node_modules/@testing-library/react/types/index.d.ts(10,31): error
TS2307: Cannot find module 'react-dom/test-utils' or its corresponding
type declarations.
```

### Problem description:

`@testing-library/react` declares `@types/react-dom` as a [`devDependency`](https://github.com/testing-library/react-testing-library/blob/8f17a2bc4fa9c2c17a623b38ded9ab18b086e457/package.json#L52), but uses it in [its external interface](https://github.com/testing-library/react-testing-library/blob/8f17a2bc4fa9c2c17a623b38ded9ab18b086e457/types/index.d.ts#L9) when used from TypeScript.

### Suggested solution:

Mark `@types/react-dom` as a regular dependency of `@testing-library/react` so that packages depending on `@testing-library/react` also pick up this dependency.

fmeum:
`@types/react-dom` is used in `types/index.d.ts`, which is an external interface, and thus needs to be declared as a regular rather than a dev dependency so that it is picked up by users of `@testing-library/react`.

**What**:

This fixes #1000.

**Why**:

`@types/react-dom` is used in `types/index.d.ts`, which is an external interface, and thus needs to be declared as a regular rather than a dev dependency so that it is picked up by users of `@testing-library/react`.

**How**:

Declare `@types/react-dom` as a regular rather than a dev dependency.

**Checklist**:

- [ ] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,11 +45,11 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.12.5",
-    "@testing-library/dom": "^8.0.0"
+    "@testing-library/dom": "^8.0.0",
+    "@types/react-dom": "^17.0.0"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.11.6",
-    "@types/react-dom": "^17.0.0",
     "dotenv-cli": "^4.0.0",
     "kcd-scripts": "^11.1.0",
     "npm-run-all": "^4.1.5",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   "dependencies": {
     "@babel/runtime": "^7.12.5",
     "@testing-library/dom": "^8.0.0",
-    "@types/react-dom": "^17.0.0"
+    "@types/react-dom": "*"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.11.6",


MatanBobi:
I'm not sure how and if we should tackle this one.
The reason is that we don't want to add 3rd party types into our bundle since not all of our users are using TypeScript.
The second reason is that having versioned types isn't good for us because we support multiple React versions (Same as we have peer deps defined as `react: *, react-dom: *`
I think that once people do have React and React-DOM and are using TypeScript, they will also have the types installed.
I'm not too familiar with the build system that you're using, but do you have `@types/react-dom` installed and it still gives you the error?

fmeum:
I'm not very familiar with the resolution mechanism for dependencies, but yarn (see e.g. https://github.com/yarnpkg/yarn/issues/3951) seems to recommend a peer dependency on `*` or a suitable version range for `@types`. Would that help address your second point?

Regarding the build system: Yes, `@types/react-dom` is installed, but the build system ensures that only packages that declare a (non-dev) dependeny on it in their `package.json` get to see it. I don't want to break or complicate the build with the popular build systems though, so if there is not good solution to this, I will use a manual workaround.

package.json
```diff
@@ -45,11 +45,11 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.12.5",
-    "@testing-library/dom": "^8.0.0"
+    "@testing-library/dom": "^8.0.0",
+    "@types/react-dom": "^17.0.0"
```
eps1lon, line 49:
```suggestion
    "@types/react-dom": "*"
```
This matches how `@types/*` packages declare their type dependencies. We definitely want to add16.9 here since this library can be used with 16.x. And 18.x will also work. But I'd just go with `*` for consistency.

fmeum:
@eps1lon Accepted your suggestion, thanks.

fmeum:
@eps1lon Friendly ping

eps1lon:
> @eps1lon Friendly ping

Could you give permissions to maintainers to edit this PR (should be on th right side next to labels, etc)? Otherwise I can't force GH actions to run on this PR.

fmeum:
> > @eps1lon Friendly ping
> 
> Could you give permissions to maintainers to edit this PR (should be on th right side next to labels, etc)? Otherwise I can't force GH actions to run on this PR.

I had it enabled. Just disabled and reenabled it to make sure it works.

eps1lon:
> > > @eps1lon Friendly ping
> > 
> > 
> > Could you give permissions to maintainers to edit this PR (should be on th right side next to labels, etc)? Otherwise I can't force GH actions to run on this PR.
> 
> I had it enabled. Just disabled and reenabled it to make sure it works.

Mkay. I guess it's GitHub codespaces being useless and not allowing me to push.

eps1lon:
@all-contributors add @fmeum for code, bug

Issue #1061, opened 2022-05-03
The current testing environment is not configured to support act(...) with vitest and React 18


MarkLyck:
"@testing-library/jest-dom": "5.16.4",
"@testing-library/react": "13.1.1",
DOM Environment: jsdom or happy-dom (happens with both)

### Problem description:

After upgrading to @testing-library/react@13.x and react@18.x with `vitest`. A lot of react tests throw the following warning:

```
Warning: The current testing environment is not configured to support act(...)
```

I noticed this after switching from jest to vitest, and not changing anything else. (The tests do not show this warning when using jest, but they they show it when using vitest in the same project).

I previously had an unrelated issue causing the same warning with Jest: https://github.com/testing-library/react-testing-library/issues/1057

I resolved that, by removing a bad implementation of `act()`. However this time the tests are identical, and works with 1 test-runner and not with another.

### Reproduction:
Credit to @[stuarthallows](https://github.com/stuarthallows) for creating a repo that reproduces the problem: https://github.com/stuarthallows/vitest-error

@stuarthallows and @dar5hak also both seem to be having the same problem as me (see my other issue here: https://github.com/testing-library/react-testing-library/issues/1057 for more details).

### Additional information.

In my case, this is breaking unit tests completely because it logs thousands of these warnings and it just stops after around 10 tests are completed.

For me a lot of the errors start with `rc-motion`

```
Warning: The current testing environment is not configured to support act(...)
    at Align (/Users/marklyck/colony/colony-frontend/node_modules/.pnpm/rc-align@4.0.12_ef5jwxihqo6n7gxfmzogljlgcm/node_modules/rc-align/lib/Align.js:45:23)
    at DomWrapper (/Users/marklyck/colony/colony-frontend/node_modules/.pnpm/rc-motion@2.5.1_ef5jwxihqo6n7gxfmzogljlgcm/node_modules/rc-motion/lib/DomWrapper.js:28:34)
    at /Users/marklyck/colony/colony-frontend/node_modules/.pnpm/rc-motion@2.5.1_ef5jwxihqo6n7gxfmzogljlgcm/node_modules/rc-motion/lib/CSSMotion.js:57:32
```

Which is probably running an animation with antd components. But I'm 99% sure that's a red-herring since the same tests and components work just fine with jest. (or testing-library/react@12 and react@17)

MarkLyck:
I found a workaround that seems to work with happy-dom (haven't tested with jsdom).

If you manually add

```
[...]

PR #1070, opened 2022-05-17
feat: Use `globalThis` if available

eps1lon:
**What**:

Closes https://github.com/testing-library/react-testing-library/issues/1061

**Why**:

Test runners like `vitest` create environments where `window` does not reflect `self`. This is not an environment that exists outside of `vitest`. This was fixed in latest versions of `vitest` (see https://github.com/vitest-dev/vitest/pull/1256).

But it's good to get in anyway to support older versions as well as other test environments that aren't configured properly but run in later versions of Node or browsers.

**How**:

Use `globalThis` if available.

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,9 @@
   },
   "eslintConfig": {
     "extends": "./node_modules/kcd-scripts/eslint.js",
+    "globals": {
+      "globalThis": "readonly"
+    },
     "rules": {
       "react/prop-types": "off",
       "react/no-adjacent-inline-elements": "off",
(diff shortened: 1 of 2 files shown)


nknapp:
I've been wondering if this should be covered by a test. This is always the hard part...

eps1lon:
We didn't cover the other branches as well. 

The specific bug can only be tested with older versions of vitest anyway.

ph-fritsche:
> I've been wondering if this should be covered by a test. This is always the hard part...

I've added a linting rule in `user-event` making sure that any global is accessed per `globalThis`.

(Because of issues with environments which shadow global variables in test. Resulting in constructors not being identical, thus errors in `jsdom` internals due to `instanceof` failing.)

See https://github.com/testing-library/user-event/pull/931

Issue #1076, opened 2022-05-23
Update the peer deps to the latest version


pedroSoaresll:
### Describe the feature you'd like:
Change the react and react-dom to latest version in peer dependency. To the new version of npm can install normally `testing-library/react` without fail and suggest the use of `--legacy-peer-deps`

### Suggested implementation:

change the version in `package.json` to the latest from both

### Describe alternatives you've considered:

An alternative is to install the packages using `--legacy-peer-dependency`

pedroSoaresll:
@eps1lon is about it

![image](https://user-images.githubusercontent.com/11558773/169869577-c0661c74-34ef-480b-84d2-0dd7ee3480fb.png)

with the new npm version peer deps are automatically installed and if found some incoherence with the versions the installation fails warning about.

eps1lon:
This error doesn't make sense to me. `^18.0.0` is fulfilled by `18.1.0`. Do you have a cloneable repository and NPM version that reproduces this problem?

jbradford:
We had the same problem locally because we'd previously used `--legacy-peer-dependency` to resolve an issue like this, when I deleted the `package-lock.json` it revealed actual packages with problems which had previously been "fixed" using `--legacy-peer-depencency`.

pedroSoaresll:
> This error doesn't make sense to me. `^18.0.0` is fulfilled by `18.1.0`. Do you have a cloneable repository and NPM version that reproduces this problem?

I'll try to replicate the problem in a individual repo and share here mentioning you.

LZL0:
I am having the same issue with the exact same output as @pedroSoaresll.

eps1lon:
Do you maybe have packages in your dependency tree hat still require React 17?

LZL0:
Yes, but I am not getting a fatal error if I remove the React Testing Library from the package.json file.

eps1lon:
Without a cloneable repository and the NPM version we won't be able to debug this issue.

morkro:
I am getting the same error message, my dependency tree is:

```json
{
  "dependencies": {
    "react": "^18.1.0",
    "react-dom": "^18.1.0"
  },
  "devDependencies": {
    "@testing-library/react": "^13.3.0"
  }
}
```

MatanBobi:
@morkro are those the only dependencies you have? Can you please share a cloneable repo?

Darkskittlz:
I'm getting the same error here 

[...]

PR #1077, opened 2022-05-23
feat: update peer deps to latest

pedroSoaresll:
**What**:

Closes the #1076 

This change is necessary because the latest version of npm fails when a peer dependency is incoherent. So to continue using the library the developers need to make use of `--legacy-peer-deps` assuming the risks themselves.

Here we can update the peer dependency and test the package guarantee for the community that all is ok 

**Why**:

**How**:

The main update is change the version of each peer deps and check if all is working as expected

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,8 +60,8 @@
     "typescript": "^4.1.2"
   },
   "peerDependencies": {
-    "react": "^18.0.0",
-    "react-dom": "^18.0.0"
+    "react": "^18.1.0",
+    "react-dom": "^18.1.0"
   },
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
We are supporting 18.0 so this would not be accurate. Let's figure https://github.com/testing-library/react-testing-library/issues/1076 out first.

eps1lon:
Closing since the underlying issue was closed

PR #1088, opened 2022-06-28
chore: mark side-effect free

jacob-ebey:
**What**:

Marking the package as side-effect free will allow bundlers such as esbuild, webpack, etc to optimize this package out of builds enabling in-source testing with tools like Vitest to run smoothly while omitting this from builds that are shipped to the browser.

**Why**:

**How**:

Simply by adding `sideEffect: false` to the package.json

**Checklist**:

- [ ] Tests N/A
- [x] Ready to be merged


nickmccurdy:
This package sets up automatic cleanup after tests, so I'm not sure if it's safe or accurate to call it side-effect free

eps1lon (changes requested):
As Nick said: This package is not side-effect free.

Even the `/pure` entrypoint isn't side-effect free since it calls a method from `@testing-library/dom` upon module evalutation.

PR #1097, opened 2022-07-26
fix: Run render queries in the render container only

Floby:
**What**: This is a bugfix (as far as I can tell).

**Why**: `render()` by default appends a new `container` to a `baseElement` which is, by default, the global variable `document.body`. This `container` ends up containing the resulting markup for the system under test. However, when using the query methods on the `RenderResult` object, the whole `document` is scanned for matches, not just the markup resulting from the `render()` call. This breaks assumptions about the scope in which queries a run : one would expect the `getByText` and other queries to be scoped to that particular render.

As a matter of fact, this is what happens when the `container` parameter is given explicitly and `baseElement` takes it value, however this is not what happens otherwise. The default behaviour (global querying on all renders) is redundant with what `screen` from `@testing-library/dom` does, without being very clear about it being global.

**How**:
+ a test reproducing the issue
+ a lot of code reading to eventually just change a single function call parameter =)

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs) ([pull request](https://github.com/testing-library/react-testing-library/pull/1097))
- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/pure.js
+++ b/src/pure.js
@@ -149,7 +149,7 @@ function renderRoot(
         return template.content
       }
     },
-    ...getQueriesForElement(baseElement, queries),
+    ...getQueriesForElement(container, queries),
   }
 }
 
--- a/src/__tests__/render.js
+++ b/src/__tests__/render.js
@@ -59,6 +59,13 @@ test('returns baseElement which defaults to document.body', () => {
   expect(baseElement).toBe(document.body)
 })
 
+test('runs queries only on its container', () => {
+  render(<p>Hello</p>)
+  const {getByText, queryByText} = render(<p>Goodbye</p>)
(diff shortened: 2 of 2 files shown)


src/pure.js
```diff
@@ -149,7 +149,7 @@ function renderRoot(
         return template.content
       }
     },
-    ...getQueriesForElement(baseElement, queries),
```
eps1lon, line 152: Could you check what other testing libraries do? This might've been intended because of portals. But the lack of tests is suspicious.

nickmccurdy:
Closing as there hasn't been activity in a year. If this is still needed someone can push a fix for the [review comment](https://github.com/testing-library/react-testing-library/pull/1097#discussion_r930281156) and ask us to reopen it.

PR #1118, opened 2022-09-04
feat(renderHook): allow passing of all render options to renderHook

TkDodo:
**What**:
allow passing of all render options to renderHook
this allows using options like `legacyRoot` as well

**Why**: As discussed here:
- https://github.com/testing-library/react-testing-library/issues/509#issuecomment-1223917350

**How**:

`renderHook` uses `render` under the hood, but not all `RenderOptions` were allowed to pass - only `wrapper` was forwarded. This PR passes all other options to `render` as well.

My intention was to allow `legacyRoot`, so if you want me to limit it to this option, I can also do that. I'm not sure if all options are worth forwarding or if there are some that don't make sense.

docs site would still need to be updated - is this a separate repo / PR ?

**Checklist**:

- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/__tests__/renderHook.js
+++ b/src/__tests__/renderHook.js
@@ -60,3 +60,29 @@ test('allows wrapper components', async () => {
 
   expect(result.current).toEqual('provided')
 })
+
+test('legacyRoot uses legacy ReactDOM.render', () => {
+  jest.spyOn(console, 'error').mockImplementation(() => {})
+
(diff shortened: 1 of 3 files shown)


eps1lon (approved):
Yep works. Thanks for the test!

eps1lon:
@all-contributors add @TkDodo for code

Issue #1051, opened 2022-04-19
React 18. set state in finally throws "act" warning, though test is passing


JoeyAtSS:
- `@testing-library/react` version: 13.1.1
- Testing Framework and version: jest 27.5.1
  
- DOM Environment: jest-dom 27.5.1
  

### Relevant code or config:

```js
  const handleSavedCardPayment = () => {
    const paymentInfo = {
      amount: formDataRef.current.amount,
      userId,
      accountId: formDataRef.current.selectedCardAccountId,
    };

    dispatch(makePayment(paymentInfo))
      .then(handlePaymentSuccess)
      .catch(noop) // already caught by thunk
      .finally(() => {
        formDataRef.current = null;
        setSubmitting(false);
      });
  };
```

### What you did:
NOTE: This error only shows up after updating to react 18 and testing library to 13.1.1, this was not an issue in earlier version.

setSubmitting toggles the button disability. When the promise is complete, it turns the re-enables the button. so the test is doing a waitFor button to be re-enabled.

the promise looks more like this

dispatch(makePayment(paymentInfo)) returns a promise, with a catch inside.

so the whole promise chain looks like this 
promisedFunction().then(() => do something).catch(error => show error). then(() => componentLevel).catch(do nothing here).finally(reset state)

### What happened:
![image](https://user-images.githubusercontent.com/96191393/163896206-a2a8d66e-fbbc-4156-be58-606d769b1060.png)

the "act" error is thrown, even though i've added a waitFor and that is passing to prove that that statement has already been rendered.

```js
await waitFor(() => expect(screen.getByRole('button', { name: /submit/i })).not.toBeDisabled());
```

### Problem description:

set state in finally seems to cause testing library to think act is incomplete, even though it is.

currently solving the error by putting the set state inside of catch
![image](https://user-images.githubusercontent.com/96191393/163897026-3357b02b-823d-4285-8ff5-c93feae3df2b.png)

However, there's the WET code, i had to put the set state into handlePaymentSuccess too, to achieve the same results.

intercaetera:
I encountered something similar while using [Formik](https://github.com/jaredpalmer/formik/), I suspect this is related. Here is a repo which reproduces this issue:

- https://github.com/intercaetera/testing-library-act

This only happens on the new version of RTL and React 18.

clothoo:
I'm using react-hook-form and experience the same act warning. Using either `waitFor` or `waitForElementToBeRemoved` doesn't remove the warning.

[...]

PR #1137, opened 2022-10-08
fix: Prevent "missing act" warning for queued microtasks

### LINKED ISSUE #1125: post-upgrade, unexpected `act` warning with `findBy` queries (includes repro)
- `@testing-library/react` version: 13.4.0
- Testing Framework and version:  jest@27.5.1
- DOM Environment: jsdom@16.7.0
- node: 16.14.2
- react: 18.2.0

### Relevant code or config:

```js
  render(<App />);
  
  //generates an unexpected act warning after upgrading
  await screen.findByText('Loading...')

  const text = await screen.findByText('testing123');
  expect(text).toBeInTheDocument();
```

### What you did:

Upgraded existing application to react 18+ and RTL 13+.

### What happened:

A number of tests starting producing `act` warnings after the upgrade.  

In this case, the component being rendered has initial "loading" state before a `fetch` is completed and the state is changed.  Our existing tests awaited the "loading" text.  This now produces an `act` warning.

### Reproduction:

I was able to reproduce the issue with a simple CRA repo, here: https://github.com/trv-wcaneira/act-warning-mcve

### Problem description:

The `act` warning is certainly a distraction over hundreds of tests, but concerning because I know it is there for a reason, and I'd like to understand what is wrong.

### Suggested solution:

I can replace the `await findBy*` queries in our tests for this use case, but I'd like to better understand why this is happening (please).  I am especially curious because after upgrading, I have noticed a number of behavior changes in existing tests (that I'll file separate issues for).

eps1lon:
**What**:

Closes https://github.com/testing-library/react-testing-library/issues/1125
Closes https://github.com/testing-library/react-testing-library/issues/1051

**Why**:

Once the `callback` passed to `waitFor` resolves, there could be other microtasks scheduled.
If we now yield back from `waitFor` we restore the `REACT_IS_ACT_ENVIRONMENT`. However, the tester `await`s the `waitFor` and before the runtime yields back to the event loop, it needs (by spec) to drain the scheduled microtasks. This may include state updates which now run without `REACT_IS_ACT_ENVIRONMENT` which triggers warning.

**How**:

Drain microtask queue before resolving `waitFor` call.

**Checklist**:

- [x] Tests
- [x] Explainer
- [x] Test in issue repro: https://github.com/trv-wcaneira/act-warning-mcve/pull/1
- [x] Test in issue repro: https://github.com/eps1lon/missing-act-setstate-finally-repro/pull/1
- ~[ ] Test with React 17~ We no longer support React 17
- [x] Test on mui/material-ui: https://github.com/mui/material-ui/pull/36221
- [x] Test in Klarna monorepo
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/end-to-end.js
+++ b/src/__tests__/end-to-end.js
@@ -1,73 +1,150 @@
 import * as React from 'react'
 import {render, waitForElementToBeRemoved, screen, waitFor} from '../'
 
-const fetchAMessage = () =>
-  new Promise(resolve => {
-    // we are using random timeout here to simulate a real-time example
-    // of an async operation calling a callback at a non-deterministic time
-    const randomTimeout = Math.floor(Math.random() * 100)
-    setTimeout(() => {
-      resolve({returnedMessage: 'Hello World'})
-    }, randomTimeout)
-  })
-
-function ComponentWithLoader() {
-  const [state, setState] = React.useState({data: undefined, loading: true})
-  React.useEffect(() => {
-    let cancelled = false
-    fetchAMessage().then(data => {
-      if (!cancelled) {
-        setState({data, loading: false})
-      }
+describe.each([
+  ['real timers', () => jest.useRealTimers()],
+  ['fake legacy timers', () => jest.useFakeTimers('legacy')],
+  ['fake modern timers', () => jest.useFakeTimers('modern')],
+])(
+  'it waits for the data to be loaded in a macrotask using %s',
+  (label, useTimers) => {
+    beforeEach(() => {
+      useTimers()
     })
 
-    return () => {
-      cancelled = true
+    afterEach(() => {
+      jest.useRealTimers()
+    })
+
+    const fetchAMessageInAMacrotask = () =>
+      new Promise(resolve => {
+        // we are using random timeout here to simulate a real-time example
+        // of an async operation calling a callback at a non-deterministic time
+        const randomTimeout = Math.floor(Math.random() * 100)
+        setTimeout(() => {
+          resolve({returnedMessage: 'Hello World'})
+        }, randomTimeout)
+      })
+
+    function ComponentWithMacrotaskLoader() {
+      const [state, setState] = React.useState({data: undefined, loading: true})
+      React.useEffect(() => {
+        let cancelled = false
+        fetchAMessageInAMacrotask().then(data => {
+          if (!cancelled) {
+            setState({data, loading: false})
+          }
+        })
+
+        return () => {
+          cancelled = true
+        }
+      }, [])
+
+      if (state.loading) {
+        return <div>Loading...</div>
+      }
+
+      return (
+        <div data-testid="message">
+          Loaded this message: {state.data.returnedMessage}!
+        </div>
+      )
     }
-  }, [])
 
-  if (state.loading) {
-    return <div>Loading...</div>
-  }
+    test('waitForElementToBeRemoved', async () => {
+      render(<ComponentWithMacrotaskLoader />)
+      const loading = () => screen.getByText('Loading...')
+      await waitForElementToBeRemoved(loading)
+      expect(screen.getByTestId('message')).toHaveTextContent(/Hello World/)
+    })
 
-  return (
-    <div data-testid="message">
-      Loaded this message: {state.data.returnedMessage}!
-    </div>
-  )
-}
+    test('waitFor', async () => {
+      render(<ComponentWithMacrotaskLoader />)
+      // eslint-disable-next-line testing-library/prefer-find-by -- Sir, this is a test.
+      await waitFor(() => screen.getByText(/Loading../))
+      // eslint-disable-next-line testing-library/prefer-find-by -- Sir, this is a test.
+      await waitFor(() => screen.getByText(/Loaded this message:/))
+      expect(screen.getByTestId('message')).toHaveTextContent(/Hello World/)
+    })
+
+    test('findBy', async () => {
+      render(<ComponentWithMacrotaskLoader />)
+      await expect(screen.findByTestId('message')).resolves.toHaveTextContent(
+        /Hello World/,
+      )
+    })
+  },
+)
 
 describe.each([
   ['real timers', () => jest.useRealTimers()],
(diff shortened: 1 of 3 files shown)


src/__tests__/end-to-end.js
```diff
@@ -1,73 +1,150 @@
...
-  }, [])
 
-  if (state.loading) {
-    return <div>Loading...</div>
-  }
+    test('waitForElementToBeRemoved', async () => {
+      render(<ComponentWithMacrotaskLoader />)
+      const loading = () => screen.getByText('Loading...')
+      await waitForElementToBeRemoved(loading)
+      expect(screen.getByTestId('message')).toHaveTextContent(/Hello World/)
```
MatanBobi, line 59: Do we have a specific reason to use `ByTestId` here instead of `ByText`? Since we advocate against using test id's, I think we can do it too :)

src/__tests__/end-to-end.js
```diff
@@ -1,73 +1,150 @@
...
+    })
+
+    test('waitFor', async () => {
+      render(<ComponentWithMacrotaskLoader />)
+      // eslint-disable-next-line testing-library/prefer-find-by -- Sir, this is a test.
```
MatanBobi, line 64: Lol, in fact I think that if `testing-library/prefer-find-by` errored against this, it's a problem because we shouldn't use `findBy` inside a `waitFor`.. Wdyt?

MatanBobi (approved):
Looks good! :)

src/__tests__/end-to-end.js
eps1lon: > Since we advocate against using test id's, I think we can do it too :)

It doesn't matter what query method we use here since they're not subject of this test.
Our tests are not production code (i.e. actual tests you write in your app). You shouldn't take these as "this is how you're supposed to write tests". I might just use `querySelector` here to make this clear since this test is not about `screen.getByTestId` but `waitForElementremoved`

src/__tests__/end-to-end.js
eps1lon: These files test Testing Library. They're not meant as documentation for "how to write tests". This is just a temporary fix. I'll probably just remove the plugin since we're not writing tests against an application/library (where these rules do make sense) but against React Testing Library itself.

The "test" I refer to here, is not the "test" you write against an app/library. The test you write against an app/library is considered "production" code from our perspective.

It's like enabling Rule of Hooks for a test that asserts on error messages if Rules of Hooks are vioalted. Or writing tests that verify runtime type-checking in TypeScript.

src/__tests__/end-to-end.js
MatanBobi: > You shouldn't take these as "this is how you're supposed to write tests".

That's not what I was trying to say, hence the approval.
Since we're using byText in other places in this test, I thought it can work better, but it's your call.

src/__tests__/end-to-end.js
MatanBobi: Again, I wasn't arguing that there's an issue with what you're doing, I was just saying that if the plugin errored in this case, it's wrong because inside a 'waitFor' we should use getBy and not findBy.

robin-drexler:
👋 @eps1lon, curious would this PR also address https://github.com/testing-library/react-testing-library/issues/1051 or this is a different issue? 🤔 

Thanks a lot!

markmssd (approved):

fastndead:
Hey! is there an ETA for this fix? We've ran into the bunch `not wrapped in act()` warnings after React 18/RLT 13 update, too many to spot -fix, but these changes actually reduce the amount of failed test by 65%, so I'm really looking forward to this being released

heath-pack:
Would love to get this in soon to try and reduce the noise from our specs, then I can see if we have any "actual" act warnings left

diogoredin:
Hey @eps1lon, any chance this going to get merged soon? We are also facing some unexpected failing tests & act warnings after moving to RTL 13 (using the legacy root still). Thank you for your work 🙏

eps1lon:
> Hey! is there an ETA for this fix? We've ran into the bunch not wrapped in act() warnings after React 18/RLT 13 update, too many to spot -fix, but these changes actually reduce the amount of failed test by 65%, so I'm really looking forward to this being released

This is very helpful information, thank you!

We have a new major planned that will include some small breaking changes from /dom. I was hesitant to merge this PR but including it in a new major makes it a lot safer to land. 

I'll clean this up, test on some larger repos and check how it performs.

src/__tests__/end-to-end.js
eps1lon: > it's wrong because inside a waitFor we should use getBy and not findBy.

That's not what the warning is suggesting though. It wants me to use `findBy` instead of `waitFor` + `getBy`. But it's not relevant because I specifically want to test that combination. It's kind of like a test for internals so there's no reason to apply production-level lint rules.

It's like linting against using the `arguments` object when my test specifically wants to test interactions with `arguments`.

MatanBobi: > It's kind of like a test for internals so there's no reason to apply production-level lint rules

I'm 100% on board with you.
I'm resolving this one :)

src/pure.js
```diff
@@ -23,7 +37,21 @@ configureDTL({
...
+      // The caller would have no chance to wrap the in-flight Promises in `act()`
+      await new Promise(resolve => {
+        setTimeout(() => {
+          resolve()
+        }, 0)
+
+        if (jestFakeTimersAreEnabled()) {
+          jest.advanceTimersByTime(0)
+        }
+      })
```
jsnajdr, line 52: An observation: if I comment out this new `await` code that drains the microtask queue, all the `end-to-end.js` tests are still passing. Both macrotask and microtask, in all timer variants.

Seems the tests would need to be amended somehow to really reproduce the bug.

eps1lon: Added tests in https://github.com/testing-library/react-testing-library/pull/1215

MatanBobi:
@ak-beam this fix was only merged to alpha and isn't merged to master yet.
Having said that, I don't think this will be backported because it uses `IS_REACT_ACT_ENVIRONMENT` which is only available in React 18.

Issue #1141, opened 2022-10-13
`renderHook` `wrapper` function is no longer called with `initialProps`


louis-young:
- `@testing-library/react` version: 13.4.0
- Testing Framework and version: Jest version 27.5.1
- DOM Environment: Jest DOM version 5.16.5 with JSDOM version 27.5.1

### Relevant code or config:

```js
import { renderHook, act } from "@testing-library/react";
import { TestContextProvider, useTestContext } from "./application";

const Wrapper = ({ children, dependency }) => {
  return (
    <TestContextProvider dependency={dependency}>
      {children}
    </TestContextProvider>
  );
};

describe("useTestContext", () => {
  it("calls `dependency` when `performSomeOperation` is called", async () => {
    const dependency = jest.fn();

    const { result } = renderHook(useTestContext, {
      wrapper: Wrapper,
      initialProps: {
        dependency,
      },
    });

    await act(async () => {
      result.current.performSomeOperation();
    });

    expect(dependency).toHaveBeenCalledTimes(1);
  });
});
```

### What you did:

I replaced `@testing-library/react-hooks` with the latest major version of `@testing-library/react`.

### What happened:

The test failed.

```
useTestContext › calls `dependency` when `performSomeOperation` is called

TypeError: dependency is not a function
```

### Reproduction:

https://codesandbox.io/s/github/louis-young/react-testing-library-issue?file=/src/index.test.js

### Problem description:

I'm under the impression that the `renderHook` function that `@testing-library/react` exports is designed to be a direct replacement for the `renderHook` function that  `@testing-library/react-hooks` exports. This doesn't appear to be the case.

### Suggested solution:

If this is a regression, I suggest that the `wrapper` option function is called with the `initialProps` values as it was in `@testing-library/react-hooks`. If this isn't a regression, I suggest that (if there isn't already, although I can't seem to find it) there's a documented idiomatic way of providing dependencies to render wrapper functions. 

Thanks 🙂

eps1lon:
So they passed the `initialProps` both to the wrapper component and the hook itself? 

I would rather avoid that API choice. `render` also has no wrapper props so `render` and `renderHook` are consistent which is more important than `renderHook` from this library being consistent with `renderHook` from another library.

no-yan:
> I would rather avoid that API choice.

Removing wrapper props from API would be okay,  since hooks with context can be tested without wrapper props.
```
test('allows wrapper components2', async () => {
[...]

PR #1143, opened 2022-10-16
fix: Pass initialProps to wrapper component

no-yan:
**What**:

This PR enables passing initialProps to wrapper component, as react-hooks-testing-library does. See testing-library/testing-library-docs#1161.

**Why**:
To test hooks which interact with context. 
For testing that, the easiest way is to create wrapper `(val) => <context.Provider value={val}` and test by changing the value of wrapper props.

**How**:
Changed renderHook logic to render wrapper component if wrapper is provided, and pass initialProps there.
Alternatively, we can do the same thing by changing `renderRoot`, but it is a bit complex (See diff in https://github.com/no-yan/react-testing-library/commit/730fe50de69197939a8b8b3ac6b82abb82818e11) and I don't think it solves the undermentioned problem.

**Checklist**:

- [ ] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs)
- [x] Tests
~~- [ ]~~ TypeScript definitions updated
- [ ] Ready to be merged N/A
      

**Note**

For compatibility with react hooks testing library([code](https://github.com/testing-library/react-hooks-testing-library/blob/121344370b2285dc3d5166a16ead71186bfe85d8/src/helpers/createTestHarness.tsx#L30-L33), [doc](https://react-hooks-testing-library.com/usage/advanced-hooks#context:~:text=Note%20the%20initialProps%20and%20the%20new%20props%20of%20rerender%20are%20also%20accessed%20by%20the%20callback%20function%20of%20the%20renderHook%20which%20the%20wrapper%20is%20provided%20to.)), I intentionally pass props to both the wrapper component and the Test Component.
I don't think there is any rationale for this specification in itself, but given that the React Testing Library is essential for React 18 migration, I think we should try to make the migration as easy as React. Therefore I implemented in a way similar to React hooks testing library.

However, this specification itself has its own limitation and will need to be changed in the future, as it does not allow to test hooks that use both props and context, like:

```
const {result} = renderHook((price) => {
    const taxRate = useContext(TaxRate)

    return price * taxRate
  })
```

### DIFF
--- a/src/__tests__/renderHook.js
+++ b/src/__tests__/renderHook.js
@@ -80,6 +80,29 @@ test('passes initialProps to a wrapper component', async () => {
   expect(result.current).toEqual('provided')
 })
 
+test('rerenders wrapper with given props', async () => {
+  const Context = React.createContext('default')
+  function Wrapper({value, children}) {
+    return <Context.Provider value={value}>{children}</Context.Provider>
+  }
+  const initialProps = {value: 'initial'}
+  const {result, rerender} = renderHook(
+    () => {
+      return React.useContext(Context)
+    },
+    {
+      wrapper: Wrapper,
+      initialProps,
+    },
+  )
+
+  expect(result.current).toEqual('initial')
+
+  rerender({value: 'updated'})
+
+  expect(result.current).toEqual('updated')
+})
+
 test('legacyRoot uses legacy ReactDOM.render', () => {
   jest.spyOn(console, 'error').mockImplementation(() => {})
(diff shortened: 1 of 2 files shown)


eps1lon (changes requested):
Let's discuss the proposed API or rather the problem this is trying to fix first: https://github.com/testing-library/react-testing-library/issues/1141#issuecomment-1279920242

no-yan:
Closed as this change is intentional.

PR #1214, opened 2023-05-22
fix: Require `await act(...)`

### LINKED ISSUE #1216: Receiving 'event was not wrapped in act' warning after upgrading to React 18
- `@testing-library/react` version: 14.0.0
- Testing Framework and version: jest
  
- DOM Environment:
  

### Relevant code or config:

```ts
// CompA.tsx

const CompA = () => null;
export default CompA;

// CompB.tsx

import { lazy } from "react";

const CompB = () => {
  const Example = lazy(() => import("./CompA"));

  return <Example />;
};

export default CompB;

// CompC.tsx

import CompB from './CompB';

const CompC = () => <CompB />;

export default CompC;

// CompC.test.tsx

import { render } from "@testing-library/react";
import CompC from './CompC';

// This test will throw `act()` warning even though CompC is not using `lazy` directly.
describe("CompC", () => {
  test(`Renders`, () => {
    expect(() => {
      render(<CompC />);
    }).not.toThrow();
  });
});
```

### What you did:

Run test using `yarn test` from a CRA configured project.

### What happened:

Receiving warning,

```
console.error
    Warning: A suspended resource finished loading inside a test, but the event was not wrapped in act(...).
    
    When testing, code that resolves suspended data should be wrapped into act(...):
    
    act(() => {
      /* finish loading suspended data */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
```

### Reproduction:

Sandbox: https://codesandbox.io/p/sandbox/demo-act-warning-nested-1p4vjn?selection=%5B%7B%22endColumn%22%3A1%2C%22endLineNumber%22%3A1%2C%22startColumn%22%3A1%2C%22startLineNumber%22%3A1%7D%5D&file=%2Fsrc%2FCompC.test.tsx%3A7%2C25

### Problem description:

`act()` warning started to appear after upgrading to React 18.

Wrapping with `waitFor` approach (as suggested [here](https://github.com/testing-library/testing-library-docs/issues/1255#issuecomment-1557558510)) has the following drawback.

Considering `lazy` is used on a shared component, we'll have to update all the tests for all the components that are using the shared component (even though those components are not using `lazy` directly). For a large codebase, we'll have to update tests for hundreds of components. Even if we have only one shared component using lazy, we'll have to update all the dependent component tests.

Seems like a lot of work!

Could we somehow keep the behavior identical to that on React 16 (`act` warning is not raised without needing `waitFor`)? It'll save a lot of developer effort.

### Suggested solution:

No warning message

### LINKED ISSUE #1385: `render` does not `await act()`
- `@testing-library/react` version: 16.2.0
- Testing Framework and version: vitest
  
- DOM Environment: chromium, via `@vitest/browser`
  

### Relevant code or config:

```tsx
import { use } from 'react';
import { render, screen } from '@testing-library/react';

function TestComponent({ promise }: { promise: Promise<unknown> }) {
  use(promise);

  return 'test';
}

test('render', async () => {
  const promise = new Promise((resolve) => {
    setTimeout(resolve, 10);
  });

  render(<TestComponent promise={promise} />);

  await expect(screen.findByText('test')).resolves.toBeInTheDocument();
});
```

### What you did:

I use `use(promise)` in components.

### What happened:

I get the following error when rendering via `@testing-library/react`'s `render()`.
```
A component suspended inside an `act` scope, but the `act` call was not awaited. When testing React components that depend on asynchronous data, you must await the result:

await act(() => ...)
```
The test also fails:
```
Caused by: TestingLibraryElementError: Unable to find an element with the text: test. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
```

### Reproduction:

See "Relevant code or config" above.

Adding an un-awaited `act` manually does not resolve the issue either:
```tsx
import { use } from 'react';
import { act, render, screen } from '@testing-library/react';

function TestComponent({ promise }: { promise: Promise<unknown> }) {
  use(promise);

  return 'test';
}

test('render', async () => {
  const promise = new Promise((resolve) => {
    setTimeout(resolve, 10);
  });

  act(() => {
    render(<TestComponent promise={promise} />);
  });

  await expect(screen.findByText('test')).resolves.toBeInTheDocument();
});
```

### Problem description:

I should never get any act warnings when using `@testing-library/react` functions.

### Suggested solution:

As the warning explicitly advises, awaiting the `act` call fixes the warning/test:
```tsx
import { use } from 'react';
import { act, render, screen } from '@testing-library/react';

function TestComponent({ promise }: { promise: Promise<unknown> }) {
  use(promise);

  return 'test';
}

test('render', async () => {
  const promise = new Promise((resolve) => {
    setTimeout(resolve, 10);
  });

  await act(() => {
    render(<TestComponent promise={promise} />);
  });

  await expect(screen.findByText('test')).resolves.toBeInTheDocument();
});
```

AFAICT, in React 19, `act` always returns a promise, so it may be best to always `await` it.
Fixing the issue would make `render` async, which may be a breaking change.
https://github.com/testing-library/react-testing-library/blob/main/src/pure.js#L149

eps1lon:
Will be released in the next alpha to test at Meta and Vercel first.

**BREAKING CHANGE**

`act` and APIs using it (e.g. `render` or `renderHook`) are now async and must be awaited. 
This change can be codemodded with https://github.com/eps1lon/codemod-missing-await-act. Current version of `user-event` is not compatible yet until `user-event` awaits all calls to `eventWrapper`.

Requires https://github.com/testing-library/dom-testing-library/pull/1369

follow-up:

- [ ] update reference docs (and write FAQ for "How to test loading states"

## fixes to missing act warnings

Sometimes you wanted to assert on a given commit of the component without caring about scheduled tasks. This always worked since these scheduled tasks were on the macrotask queue (e.g. `setTimeout`) because we automatically unmounted the component after the test ended.
However, when these tasks where on the microtasks queue (e.g. a promise that just resolved), you would get a missing act warning between the test ending and our automatic cleanup running. 

The only solution would've been to add an explicit unmount to **each** test.

Now that we require `await act(...)` these issues will no longer persist since we'll also flush the state updates on the microtask queue. If you want to continue asserting on the intermediate states, you can use `globalThis.IS_REACT_ACT_ENVIRONMENT = false`. Keep in mind though that while this flag is set, you have to manually wait for every state update to be applied e.g. by using `findBy` or `waitFor`.

## Timer changes after module initialization

Drops support for changing timers after module initialization.
This affects tests that changed the timer implementation (i.e. fake vs real timers) after modules were imported. For example, in the following test, updates are no longer flushed.

```tsx
import { render, screen } from "@testing-library/react";
import { useState } from "react";

jest.useFakeTimers;

test("updates", async () => {
  function Component() {
    const [state, setState] = useState("initial");
    useEffect(() => {
      const timeoutID = setTimeout(() => {
        setState("committed");
      }, 50);

      return () => {
        clearTimeout(timeoutID);
      };
    }, []);
    return state;
  }
  const { container } = await render(<Component />);

  // will timeout
  await waitFor(() => {
    expect(container.textContent).toEqual("committed");
  });
});
```

Instead, you have to decide on timers **before** you import modules. In Jest, you can choose timers at the config level, or keep using `jest.useFaketimers()` but call it before any `import` or `require`.

This only ever worked because we cheated a bit and flipped silently to an act environment during `waitFor`. However, since we now flush a bit more when using `IS_REACT_ACT_ENVIRONMENT`, we need to offer APIs to assert on intermediate states e.g. when your component initially displays a loading spinner that goes away in the next microtask. You can do this by simply setting `globalThis.IS_REACT_ACT_ENVIRONMENT = false` and continue to use existing APIs. React will fallback to standard (real or fake) timers and not the act queue.

Fixes https://github.com/testing-library/react-testing-library/issues/1385

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
       "react/no-adjacent-inline-elements": "off",
       "import/no-unassigned-import": "off",
       "import/named": "off",
+      "testing-library/no-await-sync-events": "off",
       "testing-library/no-container": "off",
       "testing-library/no-debugging-utils": "off",
       "testing-library/no-dom-import": "off",
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "@babel/runtime": "^7.12.5"
   },
   "devDependencies": {
-    "@testing-library/dom": "^10.0.0",
+    "@testing-library/dom": "https://pkg.csb.dev/testing-library/dom-testing-library/commit/8abda5aa/@testing-library/dom",
     "@testing-library/jest-dom": "^5.11.6",
     "@types/react": "^19.0.0",
     "@types/react-dom": "^19.0.0",
@@ -63,7 +63,7 @@
     "typescript": "^5.9.3"
   },
   "peerDependencies": {
-    "@testing-library/dom": "^10.0.0",
+    "@testing-library/dom": "^11.0.0",
     "@types/react": "^18.0.0 || ^19.0.0",
     "@types/react-dom": "^18.0.0 || ^19.0.0",
     "react": "^18.0.0 || ^19.0.0",
--- a/src/pure.js
+++ b/src/pure.js
@@ -32,18 +32,14 @@ async function waitForMicrotasks() {
 
 configureDTL({
   unstable_advanceTimersWrapper: async scope => {
-    if (getIsReactActEnvironment()) {
-      return actIfEnabled(scope)
-    } else {
-      const result = await scope()
-      await waitForMicrotasks()
-      return result
-    }
+    const result = await scope()
+    await waitForMicrotasks()
+    return result
   },
-  // We just want to run `waitFor` without IS_REACT_ACT_ENVIRONMENT
-  // But that's not necessarily how `asyncWrapper` is used since it's a public method.
-  // Let's just hope nobody else is using it.
-  asyncWrapper: async cb => {
+  eventWrapper: actIfEnabled,
+  asyncWrapper: actIfEnabled,
+  // Run `waitFor` without IS_REACT_ACT_ENVIRONMENT
+  waitForWrapper: async cb => {
     const previousActEnvironment = getIsReactActEnvironment()
     setIsReactActEnvironment(false)
     try {
@@ -52,7 +48,6 @@ configureDTL({
       setIsReactActEnvironment(previousActEnvironment)
     }
   },
-  eventWrapper: actIfEnabled,
 })
 
 // Ideally we'd just use a WeakMap where containers are keys and roots are values.
--- a/src/pure.js
+++ b/src/pure.js
@@ -1,69 +1,58 @@
 import * as React from 'react'
 import ReactDOM from 'react-dom'
+import * as DeprecatedReactTestUtils from 'react-dom/test-utils'
 import * as ReactDOMClient from 'react-dom/client'
 import {
   getQueriesForElement,
   prettyDOM,
   configure as configureDTL,
 } from '@testing-library/dom'
-import act, {
+import {
+  actIfEnabled,
   getIsReactActEnvironment,
-  setReactActEnvironment,
+  setIsReactActEnvironment,
 } from './act-compat'
 import {fireEvent} from './fire-event'
 import {getConfig, configure} from './config'
 
-function jestFakeTimersAreEnabled() {
-  /* istanbul ignore else */
-  if (typeof jest !== 'undefined' && jest !== null) {
-    return (
-      // legacy timers
-      setTimeout._isMockFunction === true || // modern timers
-      // eslint-disable-next-line prefer-object-has-own -- No Object.hasOwn in all target environments we support.
-      Object.prototype.hasOwnProperty.call(setTimeout, 'clock')
-    )
-  } // istanbul ignore next
+function enqueueTask(task) {
+  const channel = new MessageChannel()
+  channel.port1.onmessage = () => {
+    channel.port1.close()
+    task()
+  }
+  channel.port2.postMessage(undefined)
+}
 
-  return false
+async function waitForMicrotasks() {
+  return new Promise(resolve => {
+    enqueueTask(() => resolve())
+  })
 }
 
 configureDTL({
-  unstable_advanceTimersWrapper: cb => {
-    return act(cb)
+  unstable_advanceTimersWrapper: async scope => {
+    if (getIsReactActEnvironment()) {
+      return actIfEnabled(scope)
+    } else {
+      const result = await scope()
+      await waitForMicrotasks()
+      return result
+    }
   },
   // We just want to run `waitFor` without IS_REACT_ACT_ENVIRONMENT
   // But that's not necessarily how `asyncWrapper` is used since it's a public method.
   // Let's just hope nobody else is using it.
   asyncWrapper: async cb => {
     const previousActEnvironment = getIsReactActEnvironment()
-    setReactActEnvironment(false)
+    setIsReactActEnvironment(false)
     try {
-      const result = await cb()
-      // Drain microtask queue.
-      // Otherwise we'll restore the previous act() environment, before we resolve the `waitFor` call.
-      // The caller would have no chance to wrap the in-flight Promises in `act()`
-      await new Promise(resolve => {
-        setTimeout(() => {
-          resolve()
-        }, 0)
-
-        if (jestFakeTimersAreEnabled()) {
-          jest.advanceTimersByTime(0)
-        }
-      })
-
-      return result
+      return await cb()
     } finally {
-      setReactActEnvironment(previousActEnvironment)
+      setIsReactActEnvironment(previousActEnvironment)
     }
   },
-  eventWrapper: cb => {
-    let result
-    act(() => {
-      result = cb()
-    })
-    return result
-  },
+  eventWrapper: actIfEnabled,
 })
 
 // Ideally we'd just use a WeakMap where containers are keys and roots are values.
@@ -89,7 +78,7 @@ function wrapUiIfNeeded(innerElement, wrapperComponent) {
     : innerElement
 }
 
-function createConcurrentRoot(
+async function createConcurrentRoot(
   container,
   {
     hydrate,
@@ -102,7 +91,7 @@ function createConcurrentRoot(
 ) {
   let root
   if (hydrate) {
-    act(() => {
+    await actIfEnabled(() => {
       root = ReactDOMClient.hydrateRoot(
         container,
         strictModeIfNeeded(
@@ -130,29 +119,39 @@ function createConcurrentRoot(
       // Nothing to do since hydration happens when creating the root object.
     },
     render(element) {
-      root.render(element)
+      return actIfEnabled(() => {
+        root.render(element)
+      })
     },
     unmount() {
-      root.unmount()
+      return actIfEnabled(() => {
+        root.unmount()
+      })
     },
   }
 }
 
-function createLegacyRoot(container) {
+async function createLegacyRoot(container) {
   return {
     hydrate(element) {
-      ReactDOM.hydrate(element, container)
+      return actIfEnabled(() => {
+        ReactDOM.hydrate(element, container)
+      })
     },
     render(element) {
-      ReactDOM.render(element, container)
+      return actIfEnabled(() => {
+        ReactDOM.render(element, container)
+      })
     },
     unmount() {
-      ReactDOM.unmountComponentAtNode(container)
+      return actIfEnabled(() => {
+        ReactDOM.unmountComponentAtNode(container)
+      })
     },
   }
 }
 
-function renderRoot(
+async function renderRoot(
   ui,
   {
     baseElement,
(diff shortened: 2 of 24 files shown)


kentcdodds:
I always knew this day would come 😅 Everything we do in tests really should be async.

eps1lon:
There's a reason why we should do this now and a React specific reason that why we may need to do it in the future.

For now consider this test:

```tsx
function App() {
  const [ctr, setCtr] = React.useState(0);
  async function someAsyncFunction() {
    // queue a promises to be sure they all flush
    await null;
    setCtr(1);
  }
  React.useEffect(() => {
    someAsyncFunction();
  }, []);
  return ctr;
}
act(() => {
  render(<App />, container);
});
expect(container.innerHTML).toEqual("1");
```
based on https://github.com/facebook/react/blob/a21d1475ffd7225a463f2d0c0c9b732c8dd795eb/packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js#L476-L493

That test will log missing act warnings unless you unmount the tree **before** you exit your test i.e. we'd have to tell people to unmount in each test instead of relying on a single `afterEach(cleanup)`. That's a footgun we can avoid by requiring `await act()`.

I'm testing this internally first to see how it behaves across Native and Web and to check we cover most scenarios with a codemod (because that's required to land this now). Once that's done I write out a full explainer.

kentcdodds:
I'm not sure that example is a good justification honestly. Normally if you're setting state there is some visual indication of what state you just set and I would recommend people to make an assertion based on the UI that changed from that state update which would typically result in requiring a query which is wrapped in act. Can you show me a example where that wouldn't be a solution?

kentcdodds:
I'll add that not always will you have some visual indication that some state changed, but the alternative to a visual indication is typically a function call which can be mocked and the assertion can be inside a waitFor which is also wrapped in act.

eps1lon:
We already don't let people assert between what was committed from render and what was committed from `useEffect`. If that is what we want, we need to go back and redesign act. For now, the "missing act" warning is a huge turnoff we need a solution for that scales.

kentcdodds:
True, but when we're talking about something happening asynchronously (even if triggered by useEffect) that is currently possible and understandable. I'm not convinced this is justified.

MatanBobi:
> True, but when we're talking about something happening asynchronously (even if triggered by useEffect) that is currently possible and understandable. I'm not convinced this is justified.

Just to see I understood, what we're talking about for the example @eps1lon posted is something like this?
```
function App() {
  const [ctr, setCtr] = React.useState(0);
  async function someAsyncFunction() {
    // queue a promises to be sure they all flush
    await null;
    setCtr(1);
  }
  React.useEffect(() => {
    someAsyncFunction();
  }, []);
  return ctr;
}

test("Should show 1", () => {
  act(() => {
    render(<App />, container);
  });
  await waitFor(() => {
    expect(container.innerHTML).toEqual("1");
  })
})
```

Because this seems to me like a reasonable usage of `waitFor`, the problem is that people assume they won't need to because we're wrapping all interactions (including render and unmount) with `act` that should flush all effects.

SimpleCookie:
I don't feel like I am really eligible to provide very valuable feedback. But I found this PR trying to wrap my head better around `act()` especially in combination with Promises. We just switched from CRA to Vite and had to add Jest manually, while we did we encountered a ton of new bugs. Previously we didn't really need to use `act()`, but now we have to use it everywhere on user interactions such as UserEvent.click(), type etc.

Debugging and understanding these errors is by far the biggest problem I see with testing react apps. They're always really difficult to debug, understand and solve. Sometimes the act-warnings also only appear when you are running multiple tests at once in parallel which makes it all next to impossible.

TLDR:
I like the idea of enforcing correct usage of act, because I'm struggling today.

MatanBobi:
@eps1lon do we want to move on with this by pushing it to a beta branch?

nickmccurdy:
Agreed, that would also give us time to update DTL or our supported Node versions before a major release (see https://github.com/testing-library/dom-testing-library/pull/1255).

eps1lon:
I have a bit of time start of October where I want to revisit this. So testing how good codemod coverage is. Testing react-dom and react-native tests. Checking if we want this in React core and then we advance to alpha.

I don't see this blocking Node.js 18 bump. But I would also not force the Node.js 18 bump unless it is necessary or enables some new feature.

nickmccurdy:
> But I would also not force the Node.js 18 bump unless it is necessary or enables some new feature.

It's necessary for security. Node 16 is intentionally ending support early because its version of OpenSSL is no longer supported, so we should update our Node versions as soon as we can. Fortunately we're already making progress on DTL's major release dropping Node 16 (currently in alpha), so I'd also like to start migrating RTL and our other libraries.

If this PR isn't ready by then, I'd still support releasing in another major. By then I might want to make some changes to support #1209 anyway.

kentcdodds:
> @eps1lon do we want to move on with this by pushing it to an alpha branch?

I'm going to be recording testing videos for EpicWeb.dev starting tomorrow and I would love to use the alpha for these videos to make sure I'm teaching what will be available when EpicWeb.dev is launched. So yeah, an alpha release would be great for me personally :)

nickmccurdy:
@kentcdodds Sounds like a good idea! I suggest we continue working on the other breaking changes for RTL (porting testing-library/dom-testing-library/pull/1255 and then updating the dependency on DTL) in `alpha` (we're already using that channel in DTL). @eps1lon can work on this when he's ready, but if @kentcdodds needs a build sooner, we can manually put it on a different channel.

alexkrolick (commented):
FWIW, making render async and wrapping it with act() by default is also the solution I came up with in my current project (on React 17) to get stable and consistent first assertions in tests

kentcdodds:
@eps1lon, what are the chances this will actually land in React Testing Library do you think?

I'll be recording videos tomorrow that I would love to not have to re-record. I can use smoke and mirrors to record these videos with async `render` and stuff, but if you're uncertain this will land by October then maybe I should not do that. So yeah, what do you think?

On my end, like I said I've always felt like this was inevitable we'd end up doing this anyway, so I'm in favor.

types/index.d.ts
```diff
@@ -146,18 +158,16 @@ export function renderHook<
...
- * If that's not available (older version of react) then it
- * simply calls the given callback immediately
+ * `act` for the DOM renderer
  */
-export const act: typeof reactAct extends undefined
-  ? (callback: () => void) => void
-  : typeof reactAct
+export function act<T>(scope: () => T): Promise<T>
+
+export * from '@testing-library/dom'
```
eps1lon, line 173: Add comment here and to `src/pure.js` explaining why export order is important. 

iirc we had a discussion around this with Rollup people and this was actually specified behavior: first export wins not last like you would expect if it would work like object spreading i.e. `export * from ...; export { fireEvent }` is not equivalent to `module.exports = { ...dtlExports, fireEvent }`

src/pure.js
```diff
@@ -1,68 +1,34 @@
...
   // We just want to run `waitFor` without IS_REACT_ACT_ENVIRONMENT
   // But that's not necessarily how `asyncWrapper` is used since it's a public method.
   // Let's just hope nobody else is using it.
   asyncWrapper: async cb => {
     const previousActEnvironment = getIsReactActEnvironment()
-    setReactActEnvironment(false)
+    setIsReactActEnvironment(false)
     try {
-      const result = await cb()
-      // Drain microtask queue.
```
eps1lon, line 41: Check if we added a test for this.

markmssd (approved):

Codexphere92:
hi @eps1lon! are there any updates for the status of this PR? I would love to contribute if needed. It looks like this is blocking people from updating to react 19 and there is no clear direction on what needs to be done in the meantime.

mdjastrzebski:
@eps1lon any progress on this? Users of React Native TL reported the same issue (https://github.com/callstack/react-native-testing-library/issues/1787) and I wanted to coordinate the API with the RTL.

Seems like introducing async `render`, `fireEvent`, etc is the technically sound solution. That would be a breaking change, and it's only needed if tested components use suspense. What do you think about adding `renderAsync`, `fireEventAsync` etc APIs for such cases, and leaving the original sync cases as is?

eps1lon:
> What do you think about adding renderAsync, fireEventAsync etc APIs for such cases, and leaving the original sync cases as is?

I think that creates a refactoring hazard. Every time some inner Component adds Suspense or `use`, you'd have to update from `render` to `renderAsync`. However, it just continued to work with `renderAsync` if you later removed Suspense or `use`.

There's going to be migration pain but the codemod seems promising and with agents I suspect the migration won't be that hard. That's a one-time change whereas choosing between `render` and `renderAsync` would be a choice you'd have to make all the time.

We discussed having `renderAsync` first in a minor and then making the breaking change later but that just kicks the can down the road.

package.json
```diff
@@ -48,7 +48,7 @@
     "@babel/runtime": "^7.12.5"
   },
   "devDependencies": {
-    "@testing-library/dom": "^10.0.0",
+    "@testing-library/dom": "https://pkg.csb.dev/testing-library/dom-testing-library/commit/8abda5aa/@testing-library/dom",
```
eps1lon: Needs https://github.com/testing-library/dom-testing-library/pull/1369 before we can merge this

chiptus:
is this PR still planned to be merged? thanks!

mdjastrzebski:
Yeah, this PR feels really overdue 🫣

@eps1lon are there any parts missing in this PR? Do you need any help with this? I've already released async `render` in React Native TL v14.

PR #1234, opened 2023-08-17
fix: Remove optional chaining syntax to fix esm

jacobhaile:
**What**: Removing the optional chainging "?" syntax from the index file

**Why**: The optional chaining syntax can cause compilation breaks

**How**: Break out the optional changing into an `&&` statement

**Checklist**:

- [N/A] Documentation added to the
- [N/A] Tests
- [N/A] TypeScript definitions updated
- [x] Ready to be merged

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,7 +6,10 @@ import {cleanup} from './pure'
 // this ensures that tests run in isolation from each other
 // if you don't like this then either import the `pure` module
 // or set the RTL_SKIP_AUTO_CLEANUP env variable to 'true'.
-if (typeof process === 'undefined' || !process.env?.RTL_SKIP_AUTO_CLEANUP) {
+if (
+  typeof process === 'undefined' ||
+  (process.env && !process.env.RTL_SKIP_AUTO_CLEANUP)
+) {
   // ignore teardown() in code coverage because Jest does not support it
   /* istanbul ignore else */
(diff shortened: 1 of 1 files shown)


MatanBobi:
Hi @jacobhaile!
Thanks for taking the time to open this.
Since we're targeting Node 14 as the minimal engine and optional chaining is available from Node 14, I believe this change is not needed.
Thanks again!

jacobhaile:
Thanks for the time @MatanBobi!

Does this account for esm files? I upgraded react-testing-library to v14 and I'm seeing my cypress runs impacted because of that optional chaining syntax. I'm on node v16.

MatanBobi:
That's a good question @jacobhaile, I must admit I'm not entirely on the ESM train. I'll have to look into it a bit more. Not sure I'll have the time for it soon so I'm opening this PR and if someone has an opinion, they're more than free to post it :)

nickmccurdy:
> Since we're targeting Node 14 as the minimal engine and optional chaining is available from Node 14, I believe this change is not needed.

@MatanBobi I believe this is correct.

@jacobhaile I can't reproduce this issue with Cypress. Can you please share a reproduction?

jacobhaile:
I can close out -- I no longer get the error after updating to Cypress v13

MatanBobi:
Thanks for the update @jacobhaile :)

PR #1243, opened 2023-10-10
refac: Add a dependency array for useEffect

ssi02014:
**What**:

I added a dependency array for `useEffect`.

**What**:
It would be clearer to add a dependency array, and some code already has one.

**Checklist**:

- [x] Tests
- [x] Ready to be merged


MatanBobi:
Having an empty dependency array in `useEffect` does not have the same behavior as not having a dependency array, one will run only on mount and the second will run on every re-render. I'm not sure this change is needed here.
Thanks for putting the time and effort into this, but I'm closing this PR :)

Issue #1240, opened 2023-10-06
Incorrect behavior after set injectGlobals: false in jest config


krutoo:
```
"@testing-library/react": "^14.0.0",
"jest": "^29.7.0",
"jest-environment-jsdom": "^29.7.0",
```

### Relevant code or config:

I'm trying to switch from injected globals to import needed functions from `@jest/globals`

i just add `insectGlobals: false` in `jest.config.ts`

### What you did:

I added needed imports to test file and tests fails with errors like:
```
expect(received).toHaveLength(expected)

    Expected length: 0
    Received length: 1
```

### What happened:

errors are reproduced **only if there are two or more tests**

it looks like the render result is not reset and they are rendered into one document

### Reproduction:

Here an repo with example of incorrect behavior

https://github.com/krutoo/testing-library-react-jest-no-inject-globals

krutoo:
if `injectGlobas == undefined | true` there is no problem

krutoo:
it has same behavior in Bun test

leonadler:
I encountered the same problem with vitest (`config.test.globals` `true`/`false`)

This seems to solve it in vitest - please check if something similar works in Jest / Bun test:
```ts
// vitest.config.ts
export default defineConfig({
  // ...
  setupFiles: ['src/testing/setupTests.tsx')]
});
```
```tsx
// setupTests.tsx
import { cleanup } from '@testing-library/react';
import { afterEach } from 'vitest';
afterEach(() => cleanup());
```

Jest has [setupFiles](https://jestjs.io/docs/configuration#setupfiles-array) and [setupFilesAfterEnv](https://jestjs.io/docs/configuration#setupfilesafterenv-array), I'm not sure which one you would need.

krutoo:
@leonadler thank you for workaround

i will try to use this solution bun

i think that it is need to be fixed in testing library

leonadler:
It can't really be, is what I'm saying - You might be using jest, or Bun (so you yourself are already using 2 test runners!), I am using vitest, someone else might be using mocha, ava, ...
React-Testing-Library would have to probe every test runner that exists, and would have to know which test runner you are using, which it can't - unless there is a `beforeEach` global, which you are disabling with the `injectGlobals: false`.

They could add a section about it in the readme, I would agree - but they can't detect your testing framework if you basically configure it to `canBeDetected = false`.

MatanBobi:
Hi @krutoo, thanks for opening this.
[...]

PR #1244, opened 2023-10-13
feat: add warnings when globals are missing

MatanBobi:
**What**: Adds a warning when we weren't able to setup logic through test runner globals (`afterEach`, `beforeAll`, `afterAll`).

**Why**: This was raised in https://github.com/testing-library/react-testing-library/issues/1240 and I believe it's worth adding because some stuff won't work for users so they should have a warning stating why.

**How**:

Adding `else` blocks in the relevant places.

**Checklist**:

- ~~[ ] Documentation added to the
      [docs site](https://github.com/testing-library/testing-library-docs)~~
- [ ] ~~Tests~~
- [ ] ~~TypeScript definitions updated~~
- [X] Ready to be merged


src/index.js
```diff
@@ -20,6 +20,10 @@ if (typeof process === 'undefined' || !process.env?.RTL_SKIP_AUTO_CLEANUP) {
     teardown(() => {
       cleanup()
     })
+  } else {
+    console.warn(
+      `The current test runner does not support afterEach/teardown. This means we won't be able to run automatic cleanup and you should be calling cleanup() manually.`,
```
timdeschryver, line 25:
```suggestion
      `The current test runner does not support afterEach/teardown hooks. This means we won't be able to run automatic cleanup and you should be calling cleanup() manually.`,
```

src/index.js
```diff
@@ -35,6 +39,10 @@ if (typeof process === 'undefined' || !process.env?.RTL_SKIP_AUTO_CLEANUP) {
     afterAll(() => {
       setReactActEnvironment(previousIsReactActEnvironment)
     })
+  } else {
+    console.warn(
+      'The current test runner does not support beforeAll/afterAll. This means you should be setting IS_REACT_ACT_ENVIRONMENT manually.',
```
timdeschryver, line 44:
```suggestion
      'The current test runner does not support beforeAll/afterAll hooks. This means you should be setting IS_REACT_ACT_ENVIRONMENT manually.',
```

timdeschryver (approved):
I like the idea!
I'm probably "stealing" this for Angular Testing Library :)

src/index.js
MatanBobi: Thanks!

src/index.js
MatanBobi: Thanks!

timdeschryver (approved):

timdeschryver:
Thanks @ITenthusiasm that's a valid point, we'll look into this.

src/index.js
```diff
@@ -20,6 +20,10 @@ if (typeof process === 'undefined' || !process.env?.RTL_SKIP_AUTO_CLEANUP) {
     teardown(() => {
       cleanup()
     })
+  } else {
+    console.warn(
+      `The current test runner does not support afterEach/teardown hooks. This means we won't be able to run automatic cleanup and you should be calling cleanup() manually.`,
```
eps1lon, line 25: We should add instructions on how to fix this warning i.e. import from `/pure` instead or better, ensure that you always resolve to `/pure` in your test environment.

PR #1252, opened 2023-11-16
fix: log globals warning only once

### LINKED ISSUE #1249: Provide a better experience with lifecycle hooks
@MatanBobi (and @timdeschryver, if you're still interested in Angular's implementation), it seems like this change could potentially run the risk of cluttering the logs for developers who intentionally _aren't_ using the automatic cleanup functions. (For example, I'm using `vitest`, which by default doesn't have globals enabled. But I'm sticking with Vitest's defaults intentionally.) Particularly, this warning will be logged for every single test file that tests React components.

Is it possible to only display this warning once per call to `jest`/`vitest`/etc.? For instance, is there a way that an ENV var (e.g., `*TL_GLOBAL_WARNING_LOGGED`) could be set (which only lasts the lifetime of `npx jest`, `npx vitest`, etc.)? Or maybe something else would be possible?

_Originally posted by @ITenthusiasm in https://github.com/testing-library/react-testing-library/issues/1244#issuecomment-1802111077_

MatanBobi:
**What**: Resolves #1249 

**Why**: To prevent the warning from being logged in every test suite.

**How**:

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,7 +20,8 @@ if (typeof process === 'undefined' || !process.env?.RTL_SKIP_AUTO_CLEANUP) {
     teardown(() => {
       cleanup()
     })
-  } else {
+  } else if (!process.env.RTL_AFTEREACH_WARNING_LOGGED) {
+    process.env.RTL_AFTEREACH_WARNING_LOGGED = true
     console.warn(
       `The current test runner does not support afterEach/teardown hooks. This means we won't be able to run automatic cleanup and you should be calling cleanup() manually.`,
     )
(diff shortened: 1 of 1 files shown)


timdeschryver (approved):
👍

AviVahl:
@MatanBobi 
this caused a `ReferenceError: process is not defined` in our browser tests.
please note that `typeof process === 'undefined'` also goes into that if block.

MatanBobi:
@AviVahl - Sorry about that, totally missed it.
I'll issue a fix soon :)

MatanBobi:
This change was reverted :)

PR #1255, opened 2023-11-17
fix: revert missing hooks warnings

### LINKED ISSUE #1253: cleanup warnings are fired even if cleanup is fired manually
- `@testing-library/react` version: 14.1
- Testing Framework and version: Vitest 0.34
  
- DOM Environment: jsdom
  

The v14.1 introduced a warning (#1244 ), that is yelling on my vitest based tests (notice that both warnings are fired):

image

I know about the cleanup thing and I use it manually as said by the docs, and I don't expect this warning to appear.

I've also tried the "STL_SKIP_AUTO_CLEANUP" envvar (typo in the docs, it should be RLT_), but it seems to not work, and feels weird to have to add an envvar to mute a warning

MatanBobi:
**What**: This PR reverts #1252 and #1244. Resolves #1253.

**Why**: The changes here doesn't provide much added value and should be properly documented instead.
Since we can't provide a good experience for users that are manually calling `afterEach` without globals, we prefer documenting it properly.

**How**:

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,11 +20,6 @@ if (typeof process === 'undefined' || !process.env?.RTL_SKIP_AUTO_CLEANUP) {
     teardown(() => {
       cleanup()
     })
-  } else if (!process.env.RTL_AFTEREACH_WARNING_LOGGED) {
-    process.env.RTL_AFTEREACH_WARNING_LOGGED = true
-    console.warn(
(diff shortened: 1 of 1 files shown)


timdeschryver (approved):

ITenthusiasm:
Honestly, I think this is the ideal. :sweat_smile: Thanks for this 🙏🏾 :heart:

PR #1289, opened 2024-03-19
chore: Update Codecov configuration to latest

eps1lon:
Followed https://github.com/codecov/codecov-action#usage

Updated other actions as a drive-by (though checkout should probably also be latest for maximum interop with codecov action)

codecov/codecov-action docs: https://github.com/codecov/codecov-action#usage

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -34,10 +34,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: ⬇️ Checkout repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: ⎔ Setup node
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/validate.yml
```diff
@@ -59,9 +59,11 @@ jobs:
         run: npm run validate
 
       - name: ⬆️ Upload coverage report
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v4
         with:
+          fail_ci_if_error: true 
           flags: ${{ matrix.react }}
+          token: ${{ secrets.CODECOV_TOKEN }}
```
eps1lon, line 66: This will not be available on forks but tokenless uploads are still supported for those:

> Tokenless uploading is unsupported. However, PRs made from forks to the upstream public repos will support tokenless (e.g. contributors to OS projects do not need the upstream repo's Codecov token)

-- https://github.com/codecov/codecov-action#breaking-changes

eps1lon:
Going to merge and see if this needs to land first. Then I open a ticket on codecov

PR #1270, opened 2024-01-30
Drop usage of react-dom/test-utils

eps1lon:
**BREAKING CHANGE**
The `legacyRoot` option is no longer supported now that `ReactDOM.render` has been removed from React. There is no alternative.

To ensure compatibility with React Testing Library 15, make sure you're not using the `legacyRoot` option in any of your calls to `render` from React Testing Library. 

React 18 is still supported but you will see deprecation warnings from `react-dom/test-utils`. Either suppress these warnings or switch to React 19.

PR #1294, opened 2024-04-08
feat: Add support for React 19 Canary

eps1lon:
Closes https://github.com/testing-library/react-testing-library/pull/1270

We can release React 19 support as a SemVer MINOR. Even though `legacyRoot` is no longer supported, it's not really a RTL breaking change but a React breaking change that's handled in their SemVer MAJOR release. It would only be a RTL breaking change if React would be a direct dependency not a peer dependency as it is now.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -30,7 +30,7 @@ jobs:
       fail-fast: false
       matrix:
         node: [14, 16, 18]
-        react: [latest, canary, experimental]
+        react: ['18.x', latest, canary, experimental]
     runs-on: ubuntu-latest
     steps:
       - name: ⬇️ Checkout repo
--- a/src/pure.js
+++ b/src/pure.js
@@ -207,6 +207,14 @@ function render(
     wrapper,
   } = {},
 ) {
+  if (legacyRoot && typeof ReactDOM.render !== 'function') {
+    const error = new Error(
+      '`legacyRoot: true` is not supported in this version of React. Please use React 18 instead.',
+    )
+    Error.captureStackTrace(error, render)
+    throw error
+  }
+
   if (!baseElement) {
     // default to document.body instead of documentElement to avoid output of potentially-large
     // head elements (such as JSS style blocks) in debug output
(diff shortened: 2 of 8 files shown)


.github/workflows/validate.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       fail-fast: false
       matrix:
         node: [14, 16, 18]
-        react: [latest, canary, experimental]
+        react: ['18.x', latest, canary, experimental]
```
eps1lon, line 33: Added now so that we automatically continue testing React 18.x once React 19 is released.

types/index.d.ts
```diff
@@ -73,6 +73,7 @@ export interface RenderOptions<
    */
   hydrate?: boolean
   /**
+   * Only works if used with React 18.
```
MatanBobi, line 76: Should we also add this to our docs?

MatanBobi (dismissed):

types/index.d.ts
eps1lon: yep

eps1lon: source: https://github.com/testing-library/testing-library-docs/pull/1380
preview: https://deploy-preview-1380--testing-library.netlify.app/docs/react-testing-library/api#legacyroot

src/pure.js
```diff
@@ -207,6 +207,14 @@ function render(
     wrapper,
   } = {},
 ) {
+  if (legacyRoot && typeof ReactDOM.render !== 'function') {
+    const error = new Error(
+      '`legacyRoot: true` is not supported in this version of React. Please use React 18 instead.',
```
wojtekmaj, line 212: Perhaps this message could be more encouraging update, rather than encouraging tech debt?

eps1lon: Good point. Do you have a suggestion for a better warning?

wojtekmaj: My thoughts:
* `legacyRoot` is only relevant for earlier versions of React
* If you're testing with React 19, there's a high chance you are also USING React 19, then it'd be best to just remove it
* If you're testing with React 19 but somehow you are NOT using React 19 yourself (?!), you may need help with migration.

Given that, my take:

> Warning: legacyRoot is no longer supported in React 19. If your app runs React 19 or later, you should remove this flag. If your app runs React 18 or earlier, and you need help with upgrading your app, visit https://reactjs.org/link/switch-to-createroot.

eps1lon: Makes sense: https://github.com/testing-library/react-testing-library/pull/1301. The link is not that applicable since you wouldn't need to change any callsites of `render` from RTL.

Issue #1250, opened 2023-11-15
Test layout.tsx file without having validateDOMNesting warning


mezzouhri:
### Describe the feature you'd like:

As a user of react 18 with NextJS (with app directory), I would like to test routLayout.tsx file without having this warning : `Warning: validateDOMNesting(...): <html> cannot appear as a child of <div>.
        at html
        at children`
     
```
//routLayout.tsx:

export default function RootLayout({
  children,
}: {
  children: React.ReactNode;
}) {
  return (
    <html lang="fr" data-testid="rout-layout">
      <body className={inter.className}>{children}</body>
    </html>
  );
}
```

### Suggested implementation:

Maybe by modifying the [container](https://testing-library.com/docs/react-testing-library/api#container) options 

Thank you so much for your help

eps1lon:
This is a React 19 feature. In React 18 you can't render into `document`, only hydrate.

In React 19, you'd just do `render(<App />, { container: document })`. In React 18, you'd have to prepare hydrateable HTML first and then do `render(<App />, { container: document, hydrate: true })`.

All we have to do on the RTL side is ensure we properly type `container` based on `hydrate`.

> If you don't mind the TS error, you can workaround this by doing something like this:
[stackblitz.com/edit/rtl-template-68mwxh?file=src%2FApp.test.tsx](https://stackblitz.com/edit/rtl-template-68mwxh?file=src%2FApp.test.tsx)

You shouldn't do this in React 18 since that's not supported.

MatanBobi:
> You shouldn't do this in React 18 since that's not supported.

That's why I was using the canary release in my stackblitz, the typing for the canary release support this, isn't that enough?

eps1lon:
Yeah Canary should be fine. Then all we need is https://github.com/testing-library/react-testing-library/pull/1304

PR #1304, opened 2024-04-22
fix: Don't raise TypeScript errors when hydating `document`

eps1lon:
Closes https://github.com/testing-library/react-testing-library/issues/1250

As a follow-up, we need to make `@types/react-dom` a peer dependency. Since `react-dom` is a peer, the types also need to be peer. Otherwise you would get the wrong version of `@types/react-dom` if you'd install React 19.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,5 +1,5 @@
 // TypeScript Version: 3.8
-
+import * as ReactDOMClient from 'react-dom/client'
 import {
   queries,
   Queries,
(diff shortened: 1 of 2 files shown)


MatanBobi (approved):
Do we want to merge this as a fix before merging the peer deps change?

eps1lon:
> Do we want to merge this as a fix before merging the peer deps change?

Yeah, I also want to wait a week (or until React 19 alpha comes out) before merging https://github.com/testing-library/react-testing-library/pull/1305

AviVahl:
@eps1lon we see type errors since this change:
```
node_modules/@testing-library/react/types/index.d.ts:129:52 - error TS2344: Type 'Container' does not satisfy the constraint 'Element | DocumentFragment'.
  Type 'Document | Container' is not assignable to type 'Element | DocumentFragment'.
    Type 'Document' is not assignable to type 'Element | DocumentFragment'.
      Type 'Document' is missing the following properties from type 'Element': attributes, classList, className, clientHeight, and 100 more.
        Type 'Container' is not assignable to type 'DocumentFragment'.
          Type 'Document | Container' is not assignable to type 'DocumentFragment'.
            Type 'Document' is not assignable to type 'DocumentFragment'.
              Types of property 'ownerDocument' are incompatible.
                Type 'null' is not assignable to type 'Document'.
```

can be reproduced with:
```
mkdir check-testinglib
cd check-testinglib
npm init -y
npm i react react-dom @types/react @types/react-dom typescript @testing-library/react
npx tsc --init --outDir dist --skipLibCheck false
echo "export * from '@testing-library/react';" > test.ts
npx tsc --noEmit
```

Issue #1184, opened 2023-02-28
Type conflicts in testing-library vs react


ErinGreenhalgh:
- `@testing-library/react` version: 12.1.2
- Testing Framework and version: Jest 27.4.5
- DOM Environment:
jsdom: 16.6.0
react: 17.0.2

### What you did:
My team is not yet ready to upgrade to react 18, but I'd like to keep `@testing-library/react` as up to date as possible, so I'm trying to upgrade from 12.1.2 to 12.1.5. 

### What happened:
```
yarn upgrade @testing-library/react
yarn build

Found 177 errors in 2 files.

Errors  Files
   176  node_modules/@testing-library/react/node_modules/@types/react/index.d.ts:3085
     1  node_modules/@types/react/index.d.ts:3100
```
All the errors are related to type conflicts in react vs testing-library. Almost all of them are error `TS2717`:
```
node_modules/@testing-library/react/node_modules/@types/react/index.d.ts:3273:13 - error TS2717: Subsequent property declarations must have the same type.  Property 'view' must be of type 'SVGProps<SVGViewElement>', but here has type 'SVGProps<SVGViewElement>'.

3273             view: React.SVGProps<SVGViewElement>;
                 ~~~~

  node_modules/@types/react/index.d.ts:3288:13
    3288             view: React.SVGProps<SVGViewElement>;
                     ~~~~
    'view' was also declared here.
```
There is also one duplicate identifier error:
```
node_modules/@types/react/index.d.ts:3100:14 - error TS2300: Duplicate identifier 'LibraryManagedAttributes'.

3100         type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
                  ~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/@testing-library/react/node_modules/@types/react/index.d.ts:3085:14
    3085         type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
                      ~~~~~~~~~~~~~~~~~~~~~~~~
    'LibraryManagedAttributes' was also declared here.
```

### Problem description:
I expect to be able to upgrade testing-library without type conflicts. 

### Suggested solution:
I am not sure if this issue is a bug within testing-library, or some kind of dependency collision within my setup. I have tried upgrading @types/react but this results in even more errors.

eps1lon:
[...]

Issue #1277, opened 2024-01-25
Intial props not working when rendering a hook (or I've done something wrong.)


gbatterbee:
```
    "@testing-library/dom": "^8.13.0",
    "@testing-library/jest-dom": "^6.1.3",
    "@testing-library/react": "13.3.0",
    "@testing-library/react-hooks": "^8.0.1",
    "@testing-library/user-event": "^14.4.3",
     "react": "^18.2.0",
    "react-dom": "^18.2.0",   
    "react-test-renderer: 18.1.0)"
    "`node` version: 18"
    "`pnpm` (or `yarn`) version: 7"
```
### Relevant code or config:
In the Reproduction details.
Taking the example from the docs, logged the initial props;
```typescript
const CounterStepContext = createContext(1);

export const CounterStepProvider = ({
  step,
  children,
}: {
  step: number;
  children: React.ReactNode;
}) => {
  console.log({ step });
  return (
    <CounterStepContext.Provider value={step}>
      {children}
    </CounterStepContext.Provider>
  );
};

export function useCounter(initialValue = 0) {
  const [count, setCount] = useState(initialValue);
  const step = useContext(CounterStepContext);
  const increment = useCallback(() => setCount((x) => x + step), [step]);
  const reset = useCallback(() => setCount(initialValue), [initialValue]);
  return { count, increment, reset };
}

test("should use custom step when incrementing", () => {
  const wrapper = ({
    children,
    step,
  }: {
    step: number;
    children: React.ReactNode;
  }) => <CounterStepProvider step={step}>{children}</CounterStepProvider>;
  const { result, rerender } = renderHook(() => useCounter(), {
    wrapper,
    initialProps: {
      step: 2,
    },
  });

  act(() => {
    result.current.increment();
  });

  expect(result.current.count).toBe(2);

  /**
   * Change the step value
   */
  rerender({ step: 8 });

  act(() => {
    result.current.increment();
  });

  expect(result.current.count).toBe(10);
});
```
```
  console.log
    { step: undefined }
```

### What you did:

Trying to use initialProps with the renderHook

### What happened:
Initial props are not being passed.
No obvious error message, component just isn't receiving the props.

eps1lon:
`initialProps` is for the hook, not the component. 

We should change the naming to avoid confusion.  testing-library/react-hooks passed `initialProps` to both which is also confusing because it's harder to reason about (especially from a type perspective).

PR #1305, opened 2024-04-22
feat: Move `@testing-library/dom` and `@types/react-dom`  to peer dependencies

eps1lon:
Closes https://github.com/testing-library/react-testing-library/issues/1184
Docs: https://github.com/testing-library/testing-library-docs/pull/1384

**BREAKING CHANGE**:
`@testing-library/dom` was moved to a peer dependency and needs to be explicitly installed. This reduces the chance of having conflicting versions of `@testing-library/dom` installed that frequently caused bugs when used with `@testing-library/user-event`. We will also be able to allow new versions of `@testing-library/dom` being used withou a SemVer major release of `@testing-library/react` by just widening the peer dependency.
`@types/react-dom` needs to be installed if you're typechecking files using `@testing-library/react`.

Type dependencies need to match their runtime counterpart.
If `foo` is a dependency, `@types/foo` needs to be one as well.
If `foo` is a peer dependency, `@types/foo` needs to be one as well.
This is especially apparent if the constraint on `foo` spans multiple major versions.
If we'd make `@types/foo` a direct dependency, users couldn't control which major version they get.
Package managers would pick the highest.
By moving `@types/foo` to peer dependencies, users can control which version of `@types/foo` they have.

In our case specifically, you'd get type mismatches of `@types/react-dom` if we just widen the peer dependency constraint from `^18` to `^18 | ^19`. Currently, we're already compatible with the planned React 19 features so all we'd have to do is widen the constraint on release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,10 +97,12 @@ primary guiding principle is:
 ## Installation
 
 This module is distributed via [npm][npm] which is bundled with [node][node] and
-should be installed as one of your project's `devDependencies`:
+should be installed as one of your project's `devDependencies`.  
+Starting from RTL version 16, you'll also need to install
+`@testing-library/dom`:
 
 ```
-npm install --save-dev @testing-library/react
+npm install --save-dev @testing-library/react @testing-library/dom
 ```
 
 or
@@ -108,10 +110,11 @@ or
 for installation via [yarn][yarn]
 
 ```
-yarn add --dev @testing-library/react
+yarn add --dev @testing-library/react @testing-library/dom
 ```
 
-This library has `peerDependencies` listings for `react` and `react-dom`.
+This library has `peerDependencies` listings for `react`, `react-dom` and
+starting from RTL version 16 also `@testing-library/dom`.
 
 _React Testing Library versions 13+ require React v18. If your project uses an
 older version of React, be sure to install version 12:_
--- a/package.json
+++ b/package.json
@@ -45,12 +45,12 @@
   "author": "Kent C. Dodds <me@kentcdodds.com> (https://kentcdodds.com)",
   "license": "MIT",
   "dependencies": {
-    "@babel/runtime": "^7.12.5",
-    "@testing-library/dom": "^10.0.0"
+    "@babel/runtime": "^7.12.5"
   },
   "devDependencies": {
-    "@types/react-dom": "^18.2.25",
+    "@testing-library/dom": "^10.0.0",
     "@testing-library/jest-dom": "^5.11.6",
+    "@types/react-dom": "^18.2.25",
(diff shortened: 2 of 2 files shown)


MatanBobi (approved):
Thanks.
Just a thought since we're doing this and bumping a major - lately we're having many issues with DTL version mismatches in RTL and user-event, do we maybe want to think about also move DTL to peerDeps?

eps1lon:
> do we maybe want to think about also move DTL to peerDeps?

I think so. Since DTL is stateful due to the config object, having multiple versions is problematic. 

Ideally we wouldn't have the stateful config object to begin with but that requires a large refactor that needs coordination.

We don't need to rush this in this PR though. Could you help out and prepare docs and a commit making that change?

MatanBobi:
> We don't need to rush this in this PR though. Could you help out and prepare docs and a commit making that change?

Sure thing, I'll get to it later this week.

MatanBobi:
@eps1lon Wdyt about making this change as part of a major too?
https://github.com/testing-library/react-testing-library/issues/1277

eps1lon:
> Also, here's the PR to move DTL to peer: https://github.com/testing-library/react-testing-library/pull/1309

Just push this commit to this branch. I don't want to go through the alpha branch shenanigans and just ship the major with one PR.

MatanBobi:
> Just push this commit to this branch. I don't want to go through the alpha branch shenanigans and just ship the major with one PR.

Done. I haven't fixed the conflicts though as we have more changes coming to that file.

MatanBobi:
Thanks @eps1lon. I've merged from `main` and fixed the conflicts we had so this is ready to be merged IMO.

eps1lon:
Does `feat!` do anything for semantic-release?

MatanBobi:
> Does `feat!` do anything for semantic-release?

AFAIU no, it only uses the `BREAKING CHANGE` commit message, though I'm trying to follow conventional commit messages:
https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with--to-draw-attention-to-breaking-change

You can feel free to remove it if you want.

eps1lon:
We're using semantic-release so we have to stick with what that tool understands.

ph-fritsche:
> Does `feat!` do anything for semantic-release?

If you like to switch to conventional commit, it's the default config here:
```yaml
      - name: 🚀 Release
        uses: ph-fritsche/action-release@v2
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

eps1lon:
Existing setup works for us so I'd rather not experiment.

PR #1309, opened 2024-04-24
feat!: move dtl to peerDeps

MatanBobi:
BREAKING CHANGE

**What**: As we've seen some issues lately with people having multiple versions of DTL due to `user-event` setting it as a peer and since DTL has some sort of state (the `config`), we'd want to make sure that there's only one version of DTL.

I'll update the version in the text before merging as this might be RTL version 17.
Docs to follow in another PR.


MatanBobi:
Closed in favor of https://github.com/testing-library/react-testing-library/pull/1305

Issue #1312, opened 2024-04-24
Regression: Types of property `hydrate` are incompatible in `renderHook` options


samettttt:
- `@testing-library/react` version: 15.0.4
- Testing Framework and version: vitest 1.5.0
- DOM Environment: jsdom

### Relevant code or config:

```ts
import { RenderHookOptions, renderHook } from '@testing-library/react';

export const renderHookHelper = <TProps, TResult>(
    callback: (props: TProps) => TResult,
    options: RenderHookOptions<TProps> = {},
) => {
    // ...
    // bunch of code
    // ...

    return renderHook(callback, { ...options });
};
```

### What you did:

We use the `RenderHookOptions` type for a parameter in a helper function where we call `renderHook`. 

### What happened:

<img width="1104" alt="Screenshot 2024-04-24 at 20 49 01" src="https://github.com/testing-library/react-testing-library/assets/27175290/c82bc148-e4be-43b8-9f32-83cf04a8e51c">

### Reproduction:
https://stackblitz.com/edit/rtl-template-22bvkk?file=src%2FApp.test.tsx

### Problem description:

We can't simply use the `RenderHookOptions` type in our helper function. There are workaround to this problem but I thought that maybe this change was not intentional.

Issue #1311, opened 2024-04-24
TypeScript error with skipLibCheck false


eps1lon:
@eps1lon we see type errors since this change:
```
node_modules/@testing-library/react/types/index.d.ts:129:52 - error TS2344: Type 'Container' does not satisfy the constraint 'Element | DocumentFragment'.
  Type 'Document | Container' is not assignable to type 'Element | DocumentFragment'.
    Type 'Document' is not assignable to type 'Element | DocumentFragment'.
      Type 'Document' is missing the following properties from type 'Element': attributes, classList, className, clientHeight, and 100 more.
        Type 'Container' is not assignable to type 'DocumentFragment'.
          Type 'Document | Container' is not assignable to type 'DocumentFragment'.
            Type 'Document' is not assignable to type 'DocumentFragment'.
              Types of property 'ownerDocument' are incompatible.
                Type 'null' is not assignable to type 'Document'.
```

can be reproduced with:
```
mkdir check-testinglib
cd check-testinglib
npm init -y
npm i react react-dom @types/react @types/react-dom typescript @testing-library/react
npx tsc --init --outDir dist --skipLibCheck false
echo "export * from '@testing-library/react';" > test.ts
npx tsc --noEmit
```

_Originally posted by @AviVahl in https://github.com/testing-library/react-testing-library/issues/1304#issuecomment-2074958252_

AviVahl:
Ah, I should have probably opened the issue.  Thanks for that. Much appreciated.

AviVahl:
thanks for the fix. any chance for a patch release?

MatanBobi:
@AviVahl Re-running did the trick, version 15.0.5 is released with that fix :) Sorry about that and thanks for the ping.

PR #1313, opened 2024-04-24
fix: Stop restricting `container` option based on `hydrate`

eps1lon:
Closes https://github.com/testing-library/react-testing-library/issues/1312
Closes https://github.com/testing-library/react-testing-library/issues/1311

It's really not worth it to have such extensive types just to restrict `container` properly. Especially because it makes composing `render` and `renderHook` so hard. And we do want to encourage wrapping `render` in your tests.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -43,10 +43,48 @@ export type RenderResult<
   asFragment: () => DocumentFragment
 } & {[P in keyof Q]: BoundFunction<Q[P]>}
 
-export interface BaseRenderOptions<
+/** @deprecated */
+export type BaseRenderOptions<
   Q extends Queries,
   Container extends RendererableContainer | HydrateableContainer,
   BaseElement extends Element | DocumentFragment,
+> = RenderOptions<Q, Container, BaseElement>
+
+type RendererableContainer = ReactDOMClient.Container
+type HydrateableContainer = Parameters<typeof ReactDOMClient['hydrateRoot']>[0]
+/** @deprecated */
+export interface ClientRenderOptions<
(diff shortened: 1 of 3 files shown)


types/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "../node_modules/kcd-scripts/shared-tsconfig.json",
+  "compilerOptions": {
+    "skipLibCheck": true
```
eps1lon, line 4: We never actually typechecked the `.d.ts` files before.

types/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "../node_modules/kcd-scripts/shared-tsconfig.json",
+  "compilerOptions": {
+    "skipLibCheck": false
```
eps1lon, line 4: We never actually typechecked the `.d.ts` files before.

types/index.d.ts
```diff
@@ -179,19 +171,15 @@ export interface RenderHookResult<Result, Props> {
   unmount: () => void
 }
 
-export interface BaseRenderHookOptions<
+/** @deprecated */
+export type BaseRenderHookOptions<
```
MatanBobi, line 175: Just so I'll understand, what's the motivation with changing this one from an `interface` to a `type`?

MatanBobi (approved):

types/index.d.ts
eps1lon: it's just an alias for `RenderHookOptions` whereas an `interface` would create a new type.

PR #1315, opened 2024-04-26
chore: Adjust tests to workaround 18.3.0 bug

eps1lon:
Work around https://github.com/facebook/react/issues/28915. Should be fixed in 18.3.0 but the special casing should remain in case we need to test against 18.3.0 specifically.

### DIFF
--- a/tests/shouldIgnoreConsoleError.js
+++ b/tests/shouldIgnoreConsoleError.js
@@ -36,6 +36,15 @@ module.exports = function shouldIgnoreConsoleError(format) {
         // Ignore it too.
         return true
       }
+      if (
+        format.startsWith(
+          'Warning: `ReactDOMTestUtils.act` is deprecated in favor of `React.act`.',
+        )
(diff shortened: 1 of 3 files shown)


tests/shouldIgnoreConsoleError.js
```diff
@@ -36,6 +36,15 @@ module.exports = function shouldIgnoreConsoleError(format) {
         // Ignore it too.
         return true
       }
+      if (
+        format.startsWith(
+          'Warning: `ReactDOMTestUtils.act` is deprecated in favor of `React.act`.',
+        )
+      ) {
+        // This is a React bug in 18.3.0.
+        // Versions with `ReactDOMTestUtils.ac` being deprecated, should have `React.act`
```
MatanBobi, line 45: This is really minor, sorry:
```suggestion
        // Versions with `ReactDOMTestUtils.act` being deprecated, should have `React.act`
```

Issue #1316, opened 2024-04-28
act shows up as deprecated when using react 18.3


ambarvm:
- `@testing-library/react` version: 15.0.5
- Testing Framework and version: jest@29.2.1
  
- DOM Environment: jsdom@20.0.1
  

### Relevant code or config:

```js
import { act } from '@testing-library/react';
act(() => {});
```

Act is marked as deprecated.
```
The signature (callback: () => VoidOrUndefinedOnly): void of act is deprecated.
```

### What you did:

Updated react to 18.3 in an existing project

### What happened:

act imported from testing library is marked deprecated.

### Reproduction:

https://stackblitz.com/edit/rtl-template-749zgo?file=src%2FApp.test.tsx

### Problem description:

The js code correctly exports the new `React.act` which is not deprecated. https://github.com/testing-library/react-testing-library/blob/f6a1677501b53471f6a989078726aeb0dea114be/src/act-compat.js#L5
But the types file always exports the type of deprecated act. https://github.com/testing-library/react-testing-library/blob/f6a1677501b53471f6a989078726aeb0dea114be/types/index.d.ts#L10

### Suggested solution:

Add another condition to the type definition to export React.act when it is present.
```ts
export const act: typeof reactAct extends undefined ? (typeof reactDomTestUtilsAct extends undefined
  ? (callback: () => void) => void
  : typeof reactDomTestUtilsAct) : typeof reactAct
```

I think we can also remove the case for act not being present since react peerDependency is ^18.
```ts
export const act: typeof reactAct extends undefined ? typeof reactDomTestUtilsAct : typeof reactAct
```

MatanBobi:
Thanks @ambarvm, I've created https://github.com/testing-library/react-testing-library/pull/1319 to resolve this :)

PR #1319, opened 2024-04-29
fix: export new act when available

MatanBobi:
Resolves https://github.com/testing-library/react-testing-library/issues/1316

**What**: This PR exports the correct act if available to avoid deprecation warnings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,8 +63,14 @@
     "typescript": "^4.1.2"
   },
   "peerDependencies": {
+    "@types/react": "^18.0.0",
     "react": "^18.0.0",
     "react-dom": "^18.0.0"
+  },
+  "peerDependenciesMeta": {
+    "@types/react": {
+      "optional": true
+    }
   },
   "eslintConfig": {
     "extends": "./node_modules/kcd-scripts/eslint.js",
--- a/package.json
+++ b/package.json
@@ -51,12 +51,13 @@
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^5.11.6",
+    "@types/react": "^18.3.1",
     "chalk": "^4.1.2",
     "dotenv-cli": "^4.0.0",
     "jest-diff": "^29.7.0",
     "kcd-scripts": "^13.0.0",
     "npm-run-all": "^4.1.5",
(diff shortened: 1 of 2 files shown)


eps1lon (commented):
How did you test this?

MatanBobi:
> How did you test this?

I've used `npm link` locally to see that the error is there before I link and removed after I link. Having said that, I wasn't able to install an older type of `react` and `react-dom`, for some reason I always get the same d.ts files which show act from test-utils as deprecated and act from react exported so that's odd.

From your comment, I realized that we probably also need to define `@types/react` as a peer dependency now? I saw that it's a dependency for `@types/react-dom` so I'm not sure if we want to rely on that or not.

eps1lon:
Yeah I don't think this ever worked for versions where `react-dom/test-utils` didn't have `act`.

TypeScript should error if you try to import something that doesn't exist. 

We probably need to `@ts-ignore` the `React.act` import and then check if it's `any` (`typeof act extends never ? true : false`) instead of `undefined`.

Let's double check this works in versions without `React.act` by forcing an older version of `@types/react` via `overrides` (npm) or `resolutions` (yarn, pnpm). Make sure to clear lockfile and node_modules if you use npm.

eps1lon:
And we also need `@types/react` as an optional peer dependency. We always needed that though since we already use `React.ReactNode`. It just happened to work because TypeScript always makes namespaces available globally.

MatanBobi:
Using `overrides` didn't help as I'm using `link` so the dependency is still taken locally. I've tried installing an older version of `@types/react` and `@types/react-dom` and link. Things seem to work.
I've pushed the fixes. I'll also create a PR to the docs site to state that users will probably want to install `@types/react` as it's a peer dependency now.

eps1lon (approved):
Works: https://github.com/eps1lon/rtl-deprecated-act-repro

Only errors on an error specific to older versions of `@types/react`.

mwegter95:
Yesss thank you. This helped me. When I saw your post, I downloaded the latest version and my program started running, no warning after that. Tyty

`npm install @testing-library/react@latest @testing-library/jest-dom@latest`

PR #1330, opened 2024-05-20
support `document` containers in cleanup

### LINKED ISSUE #1329: When `container` is `document`, cleanup steps fails with "TypeError: Cannot read properties of null (reading 'removeChild')"
Just opening this for tracking with the associated PR.

When:

```js
render(<Component />, { container: document });
```

(Untested: This bug _may_ require that `Component` render a `<body>` element.)

Then:

```
TypeError: Cannot read properties of null (reading 'removeChild')
```

Because:

```
// pure.js
    if (container.parentNode === document.body) {
      document.body.removeChild(container);
    }
```

If we first check that `document.body` is not null, then this error will not occur.

quisido:
**What**:

When `container` is set to `document`, the cleanup step fails:

```
TypeError: Cannot read properties of null (reading 'removeChild')
```

This is because the code assumes `document.body` exists:

```
    if (container.parentNode === document.body) {
      document.body.removeChild(container);
    }
```

In reality, both `container.parentNode` _and_ `document.body` are `null`.

**Why**:

The tests erroneously fail.

**How**:

The attempt to remove the container from the body now only occurs when the body exists.

**Checklist**:

- [ ] Tests
- [x] Ready to be merged
      

resolves #1329

### DIFF
--- a/src/pure.js
+++ b/src/pure.js
@@ -263,7 +263,7 @@ function cleanup() {
     act(() => {
       root.unmount()
     })
-    if (container.parentNode === document.body) {
+    if (document.body !== null && container.parentNode === document.body) {
       document.body.removeChild(container)
     }
   })


eps1lon:
Please add a test.

I don't understand why `document.body` would be `null`? Isn't that always defined?

quisido:
I don't know the setup, but is null when container is document. Locally, I fixed this by setting it to document.createElement('body').

I unfortunately do not have bandwidth to write a test for this, but the reproduction steps should be simple enough.

I ran into this specifically when writing a test for a NextJS root layout, which attempts to mount an HTML and body element. The docs say to use container: document when rendering your own HTML element, which fixed the previous error during test but introduced the OP error during cleanup.

MatanBobi:
@quisido thanks for taking the time to add this but we are not merging changes without proper coverage for them.

quisido:
> @quisido thanks for taking the time to add this but we are not merging changes without proper coverage for them.

Understandable. The PR is welcome to be updated. I opened an issue for it for tracking as well. I can try updating tests as time permits, but I wanted the problem and solution to both be documented.

MatanBobi:
I tried writing a test for this but it passes in our tests because `document.body` is not `null`. I'm trying to figure out what's the difference in the setup.

Hardanish-Singh (approved):

PR #1333, opened 2024-06-01
feat(pure): add renderOptions support to render

naorpeled:
**What**:

**Why**:

**How**:

**Checklist**:

- [x] Tests
- [ ] Ready to be merged
      

closes 1297

### DIFF
--- a/src/__tests__/render.js
+++ b/src/__tests__/render.js
@@ -219,6 +219,46 @@ describe('render API', () => {
     expect(wrapperComponentMountEffect).toHaveBeenCalledTimes(1)
   })
 
+  test('renderOptions are passed to createRoot', () => {
+    function Component() {
+      const id = React.useId()
+      return <div id={id} />
+    }
+
+    const container = document.createElement('div')
+    document.body.appendChild(container)
+
+    render(<Component />, {
+      container,
+      renderOptions: {
+        identifierPrefix: 'some-identifier-prefix',
+      },
+    })
+
+    expect(container.firstChild.id).toContain('some-identifier-prefix')
+  })
+
+  test('renderOptions are passed to hydrateRoot', () => {
+    function Component() {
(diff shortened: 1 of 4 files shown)


eps1lon:
We should flatten this instead of allowing arbitrary options. We don't know if a passing all through will make sense. If people want to access the root options directly, they can always switch to `createRoot` directly.

naorpeled:
> We should flatten this instead of allowing arbitrary options. We don't know if a passing all through will make sense. If people want to access the root options directly, they can always switch to `createRoot` directly.

how should we handle the cases where the consumer uses the non-canary version that currently does not expose those options?

eps1lon:
> how should we handle the cases where the consumer uses the non-canary version that currently does not expose those options?

Is there something different in that regard between your initial version and the version I proposed?

naorpeled:
> > how should we handle the cases where the consumer uses the non-canary version that currently does not expose those options?
> 
> Is there something different in that regard between your initial version and the version I proposed?

It's different in the sense that we base the type definitions on what the version the consumer uses,
in React 18 they won't get type autocompletions for v19 canary definitions, and vice verse.

eps1lon:
> Afaik it's different in the sense that we base the type definitions on what the version the consumer uses,
in React 18 they won't get type autocompletions for v19 canary definitions, and vice verse.

That's intended though, no? I don't see how type concerns need to dictate our API. It seems to me we can make flat options work. It just requires a bit more type-implementation work on our part.

eps1lon:
Alternate: https://github.com/testing-library/react-testing-library/pull/1354

naorpeled:
> Alternate: https://github.com/testing-library/react-testing-library/pull/1354

Sorry I ended up not finishing this task, had some IRL things to deal with.

Thanks for taking this task.

Issue #1338, opened 2024-06-26
RTL issue with patching asyncWrapper/eventWrapper (with act()) for userEvent API in Vitest


millerized:
We've discovered that when using latest versions of Vitest, RTL, and userEvent the userEvent APIs are not patched with act() and this results in "not wrapped in act() warnings. 

```
"@testing-library/dom": "10.2.0",
"@testing-library/react": "^16.0.0",
"@testing-library/user-event": "^14.5.2",

"jsdom": "^24.0.0",
"vite": "^5.3.1",
"vitest": "^1.6.0",
"vitest-dom": "^0.1.1"
```

DOM Environment: JSDOM

### Relevant code or config:
n/a

### What you did:
testing simple MUI component interactions with userEvent.setup() APIs.

### What happened:

All interactions result in the "not wrapped in act() warning":

```bash
stderr | src/select.spec.jsx > Select > works with user interactions
Warning: An update to ForwardRef(FormControl) inside a test was not wrapped in act(...).

When testing, code that causes React state updates should be wrapped into act(...):

act(() => {
  /* fire events that update state */
});
/* assert on the output */

```

### Reproduction:

Please see this stack blitz repro: https://stackblitz.com/edit/vitejs-vite-x8atqu?file=package.json%3AL34,src%2Fselect.spec.jsx&terminal=dev

You can rerun the test with `nam test`.
Note that the first test passes but the second one will not. I skipped the failing test because the stack trace is harder to read.

### Problem description:

IIUC RTL is supposed to "monkey patch" DTL so that all userEvent APIs are wrapped in act().

### Suggested solution:

We (cc @kiel-nokia) stepped through all the code to debug this and I'm not sure we have an easy/simple solution to present as this looks like a runtime side effects issue.

PR #1343, opened 2024-07-11
fix: esm output's extension to .mjs

syi0808:
**What**:
**Why**:
More Information: https://github.com/kentcdodds/kcd-scripts/pull/244

**How**:
Change module field extension to `.mjs` and upgrade `kcd-scripts` after [pr](https://github.com/kentcdodds/kcd-scripts/pull/244) released.

**Checklist**:

- [ ] Tests
- [ ] Ready to be merged

Resolves: #1338 testing-library/user-event#1213

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Simple and complete React DOM testing utilities that encourage good testing practices.",
   "main": "dist/index.js",
   "types": "types/index.d.ts",
-  "module": "dist/@testing-library/react.esm.js",
+  "module": "dist/@testing-library/react.esm.mjs",
   "engines": {
     "node": ">=18"
   },


syi0808:
I don't know why formatting failed in workflow. In local, It passed.

eps1lon (changes requested):
> Why:
> More Information: https://github.com/kentcdodds/kcd-scripts/pull/244

It's not clear from this issue why we need to do this. Do you have a minimal repro highlighting what this fixes? 

Using .mjs instead of .js is not trivial since older build setups may stop recognizing these files as JS so there should be a justification for making that change.

syi0808:
> > Why:
> 
> > More Information: https://github.com/kentcdodds/kcd-scripts/pull/244
> 
> 
> 
> It's not clear from this issue why we need to do this. Do you have a minimal repro highlighting what this fixes? 
> 
> 
> 
> Using .mjs instead of .js is not trivial since older build setups may stop recognizing these files as JS so there should be a justification for making that change.

To be honest, I can't logically explain why it's an issue. I just found out that if you have a .mjs extension or a package.json with type:module, this will fix it.

I checked by modifying node_modules myself, so I don't know how to share repro. Should I give it to you as a compressed file?

syi0808:
# Repro
## Project
[Archive.zip](https://github.com/user-attachments/files/16241564/Archive.zip)
After decompressing, run `npm i`
## testing library attached debug log
[@testing-library.zip](https://github.com/user-attachments/files/16241580/%40testing-library.zip)
After decompressing, paste in node_modules

syi0808:
And more detail. @eps1lon 

Testing-library/user-event in ESM format was fetching testing-library/dom in CJS format. But testing-library/react before the extension changed was getting testing-library/dom in ESM format.
After changing the extension, both are importing testing-library/dom in CJS format.

I'm not sure why you're looking for testing-library/dom in CJS format, but I think the current implementation is meaningful in unifying the different ways of working between the two libraries.

Screenshot 2024-07-16 at 9 26 28 AM
Screenshot 2024-07-16 at 9 27 06 AM

syi0808:
Can we check it? @eps1lon @MatanBobi

millerized:
> > Why:
> > More Information: [kentcdodds/kcd-scripts#244](https://github.com/kentcdodds/kcd-scripts/pull/244)
> 
> It's not clear from this issue why we need to do this. Do you have a minimal repro highlighting what this fixes?
> 
> Using .mjs instead of .js is not trivial since older build setups may stop recognizing these files as JS so there should be a justification for making that change.

Hi @eps1lon! I believe there could be an issue with how the modules are being loaded based on ESM vs. CJS semantics. As noted here https://github.com/testing-library/react-testing-library/issues/1338 using latest Vitest, RTL, and userEvent together is "broken" and displaying errors in the console. We believe this is due to the userEvent APIs not being patched by RTL at runtime. When we debug this we find that identical esm/cjs modules are loaded and the incorrect one is monkey patched by RTL and this is why we see "not wrapped in act()" warnings.

I'm not sure why @syi0808's change seems to fix this but if I had to guess it's the way that modules are loaded in node/vitest vs jest and this change makes things less ambiguous for non-jest runners. 🤷‍♂️

PR #1353, opened 2024-08-28
Fix React Canary and Experimental tests

eps1lon:
`react-dom/server` requires `MessageChannel` now. We polyfill it with the one from `worker_threads`. Don't try this at home.


MatanBobi (approved):
You made me laugh with the description 😅

PR #1365, opened 2024-11-25
Add async render APIs

eps1lon:
Not final. For internal testing for now. Alternate to https://github.com/testing-library/react-testing-library/pull/1214

- `render(...)` -> `await renderAsync(...)`
- `act(...)` -> `await actAsync(...)`
- `cleanup(...)` -> `await cleanupAsync(...)`
- `renderHook(...)` -> `await renderHookAsync(...)`


src/pure-async.js
```diff
@@ -0,0 +1,330 @@
...
+import {getConfig, configure} from './config'
+
+async function act(scope) {
+  const previousActEnvironment = getIsReactActEnvironment()
+  setReactActEnvironment(true)
+  try {
+    // React.act isn't async yet so we need to force it.
+    return await React.act(async () => {
+      scope()
+    })
```
phryneas, line 21: pretty sure this should be
```suggestion
    return await React.act(async () => {
      await scope()
    })
```
or
```suggestion
    return await React.act(async () => {
      return scope()
    })
```

eps1lon:
We'll with https://github.com/testing-library/react-testing-library/pull/1214 since this approach doesn't work for libraries like user-event that rely on `configure()`

PR #1390, opened 2025-04-02
Add `reactStrictMode` as an option to `render`

bernardobelchior:
**What**:

Add an ability to override the global `reactStrictMode` configuration on a per-test basis. 

**Why**:

At MUI, our tests use a small wrapper around React Testing Library and [had this functionality](https://github.com/mui/material-ui/blob/a80e7b7270a1af786870b7a4ad84a1d17c4ce5bd/packages-internal/test-utils/src/createRenderer.tsx#L639). However, it was broken in the upgrade to React 19 since `StrictMode` won't double call effects on initial mount unless `StrictMode` is rendered at the root of the tree. Since we were rendering `React.StrictMode` inside the component passed to the `wrapper` property, effects aren't being called twice in React 19. 

**How**:

Added a `reactStrictMode` to `renderOptions`. 

**Checklist**:

- [x] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs): https://github.com/testing-library/testing-library-docs/pull/1479
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/render.js
+++ b/src/__tests__/render.js
@@ -262,4 +262,38 @@ describe('render API', () => {
       `\`legacyRoot: true\` is not supported in this version of React. If your app runs React 19 or later, you should remove this flag. If your app runs React 18 or earlier, visit https://react.dev/blog/2022/03/08/react-18-upgrade-guide for upgrade instructions.`,
     )
   })
+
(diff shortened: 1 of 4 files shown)


eps1lon:
@all-contributors add @bernardobelchior for code, docs

PR #1396, opened 2025-04-19
Allow setting a displayName for wrapper component created in renderHook

vikr01:
**What**:
- Adding a displayName so that if you try to inspect the wrapper component `renderHook` creates, you can (i.e. `console.log`, `useFiber`, `ReactCurrentOwner`)

**Why**:
- It's a simple change and makes debugging smoother for those who need it

**How**:

**Checklist**:

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/renderHook.js
+++ b/src/__tests__/renderHook.js
@@ -53,10 +53,10 @@ test('allows rerendering', () => {
 test('allows setting a displayName', () => {
   let capturedElement = null
 
-  const spyWrapper = ({ children }) => {
+  const spyWrapper = ({children}) => {
     // Capture the hook element React creates
-    capturedElement = React.Children.only(children);
-    return <>{children}</>;
+    capturedElement = React.Children.only(children)
+    return <>{children}</>
   }
 
   const useMyLocalHook = jest.fn()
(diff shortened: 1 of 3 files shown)


MatanBobi:
Thanks for this.
My main concern with this PR is that people might use it and assert on it for some reason. What's the real benefit you get from seeing the display name here? E.g. what's the use case where you felt like you need it?

vikr01:
For hooks that use the internal fiber nodes, such as https://github.com/pmndrs/its-fine, when you check the react tree, you'd want to see a name you're familiar with otherwise it's just going to be `null`.

https://github.com/pmndrs/its-fine/blob/main/src/index.test.tsx#L146

In my specific use case, I'm creating a hook that gives you a string that is unique to the *structure* of an element, so that you can restore values based on its position (i.e. after suspending). So in this scenario, I want this actual displayName string because the pointer for the component is destroyed.

https://github.com/vikr01/react-suspense-tools/blob/initial_build/packages/useStructuralId/src/__tests__/__snapshots__/useStructuralId.test.tsx.snap#L3

So yes, it is something you would want to assert on, but you can only assert on it if you're able to reach the component by attempting to see the element tree.

eps1lon (changes requested):
The function is named so I don't understand why you'd need to override the function. `renderHook` is just a convenience helper and doesn't need all possible configuration options. Especially if it's to support libraries that rely on React internals and contribute to forking the ecosystem

If you need a special `displayName`, you can write your own wrapper.

vikr01:
That's fair, technically *it is* named.

PR #1401, opened 2025-05-28
fix: Update 'onCaughtError' type inference in 'RenderOptions' to work with React v19

aicest:
**What**: To fix incorrect `onCaughtError` type inference in `RenderOptions`.

Repro: https://codesandbox.io/p/devbox/delicate-voice-7wnf7z?file=%2Ftest%2Frender.test-d.ts%3A6%2C5-6%2C28

Screenshot 2025-05-28 at 23 40 10

**Why**: _As explained above._

**How**: Update TS types.

**Checklist**:

- [x] Tests
      - React v18 types: No impact
      - React v19 types: Fixed
- [x] Ready to be merged


aicest:
Type checking failed.

image

Solution: https://github.com/fb55/entities/pull/1851

Workaround: Set `skipLibCheck` in `tsconfig.json` to `true`

types/test.tsx
```diff
@@ -268,7 +268,7 @@ export function testErrorHandlers() {
   render(null, {
     // Should work with React 19 types
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-expect-error
+    // @ts-ignore
```
aicest, line 271: Since v19 types are installed by default and work fine with v19 types, the TS compiler reports that the `@ts-expect-error` directive is unused, and we should remove it, but for compatibility with the v18 types in tests, we finally decided to replace the directive with `@ts-ignore`...

aicest: Done, thanks for review.

eps1lon (approved):

Issue #1187, opened 2023-03-06
Support sinon fake timers


thatsmydoing:
### Describe the feature you'd like:

When using `waitFor` and sinon fake timers enabled, the test gets stuck and eventually times out. This seems to be a consequence of https://github.com/testing-library/react-testing-library/commit/f78839bf4147a777a823e33a429bcf5de9562f9e which adds a `setTimeout`.

### Suggested implementation:

Supporting it in the same way jest timers are supported is probably okay, alternatively provide configuration for generically supporting arbitrary fake timer implementations.

### Describe alternatives you've considered:

Adding the following to our test setup seems to work with the current version but it could easily get out of sync and it would be nice to have it built in or provide nicer abstraction to configure.

```
import { configure } from '@testing-library/react';
import sinon from 'sinon';

configure({
  asyncWrapper: async (cb) => {
    // our tests run only in the browser but this should more properly be getIsReactActEnvironment()
    const previousActEnvironment = window.IS_REACT_ACT_ENVIRONMENT;
    window.IS_REACT_ACT_ENVIRONMENT = false;
    try {
      const result = await cb();
      // Drain microtask queue.
      // Otherwise we'll restore the previous act() environment, before we resolve the `waitFor` call.
      // The caller would have no chance to wrap the in-flight Promises in `act()`
      await new Promise(resolve => {
        setTimeout(() => {
          resolve();
        }, 0);
        // run callbacks if we've mocked out global timers
        if (window.Date.isFake) {
          sinon.clock.tick(0);
        }
      });
      return result;
    } finally {
      window.IS_REACT_ACT_ENVIRONMENT = previousActEnvironment;
    }
  },
});
```

### Teachability, Documentation, Adoption, Migration Strategy:

Ideally, it should just work. Otherwise, examples for how to configure it would be nice.

thatsmydoing:
Alternatively, flush the queue without using `setTimeout` like what react does in https://github.com/facebook/react/blob/main/packages/shared/enqueueTask.js

eugenet8k:
This would be great, jest fake timers are supported in RTL, but sinon fake times are widely used. We can't build a few scenarios due to this limitation at the moment.

Aryan3212:
Use this snippet if you only want the current time to be faked

```js
sinon.useFakeTimers({
        now: (new Date(2024, 7, 21)).getTime(),
        toFake: ['Date'],
      });
```

PR #1443, opened 2026-03-16
fix: add Vitest fake timer detection to asyncWrapper

### LINKED ISSUE #1197: Update to v14 breaks @testing-library/user-event on Vitest

penkzhou:
**What**: Add Vitest fake timer detection to the `asyncWrapper` in `pure.js`, so that `waitFor` (and by extension `user.click` via `@testing-library/user-event`) no longer hangs when Vitest fake timers are active.

**Why**: The `asyncWrapper` uses `setTimeout(0)` to drain the microtask queue after a callback completes. When fake timers are active, this `setTimeout` never fires unless the test framework advances timers. The existing code only checks for Jest fake timers (via the `jest` global), so Vitest users with `vi.useFakeTimers()` experience indefinite hangs in `waitFor` and `user.click`. This is the root cause of #1197 and relates to #1187.

**How**:

- Added `vitestFakeTimersAreEnabled()` — checks for the `vi` global AND `setTimeout.clock` (set by `@sinonjs/fake-timers`, which Vitest uses internally). Both conditions must be true to avoid false positives when Vitest is present but real timers are in use.
- Extracted `advanceFakeTimers()` — a unified helper that tries Jest first, then Vitest, keeping the `asyncWrapper` itself clean.
- The `asyncWrapper`'s `setTimeout(0)` + advance pattern is unchanged; only the detection scope is widened.
- Added regression tests covering Jest legacy/modern fake timers, real timers, and fake timer detection logic (unit-level assertions on `setTimeout.clock` and `_isMockFunction` presence).

Note: The Vitest code path cannot be fully integration-tested within Jest (since `jest` is always available as a global and takes priority in detection). The detection logic is verified via unit tests that assert the signals each framework sets on `setTimeout`. Full Vitest integration was verified manually in a separate Vitest project.

**Checklist**:

- [ ] Documentation added to the [docs site](https://github.com/testing-library/testing-library-docs) (N/A — internal detection logic, no API change)
- [x] Tests
- [ ] TypeScript definitions updated (N/A — no public API change)
- [x] Ready to be merged

Closes #1197
Relates to #1187


penkzhou:
I encountered this leak during development. Since the source of the leak is within this specific library, I would like to have it fixed so that more people can benefit from the solution. cc @eps1lon @MichaelDeBoey

Ziiyodullayevv (commented):
Important fix for Vitest compatibility. Adding fake timer detection to asyncWrapper ensures proper async handling in Vitest test environments.

src/pure.js
```diff
@@ -27,6 +27,32 @@ function jestFakeTimersAreEnabled() {
   return false
 }
 
+/* istanbul ignore next */
+function vitestFakeTimersAreEnabled() {
+  // Vitest uses @sinonjs/fake-timers which attaches a `clock` property to setTimeout.
+  // Unlike Jest, Vitest does not set `setTimeout._isMockFunction`.
+  return (
+    // eslint-disable-next-line no-undef -- `vi` is a global in Vitest, like `jest` in Jest.
+    typeof vi !== 'undefined' &&
```
snowystinger, line 36: Isn't `vi` non-existent by default? https://v1.vitest.dev/config/#globals

snowystinger (commented):
It'd be really nice if we could expose some sort of property so that we don't need to do fake timer detection in other libraries. For instance, it'd be nice in user event to auto advance a sequence of timers for a series of user actions when fake timers are in use.
https://github.com/testing-library/user-event/pull/1324

From what I can tell right now, we'd be doing a similar `if (jest/vi/setTimeout.clock.tick)` which could diverge from the tests here

PR #1461, opened 2026-04-22
fix(config): preserve reactStrictMode when only DTL options are configured

Bojun-Vvibe:
**Repo:** testing-library/react-testing-library (⭐ 19000)
**Type:** bugfix
**Files changed:** 2
**Lines:** +10/-3

## What
In `src/config.js`, `configure()` destructured `reactStrictMode` from the
incoming partial config and unconditionally wrote it back into the RTL
config store. When callers passed only DTL options (e.g.
`configure({testIdAttribute: 'foo'})`), `reactStrictMode` was destructured
as `undefined` and silently clobbered any previously set value. The fix
only updates `reactStrictMode` when the caller actually provided it, and
adds a regression test in `src/__tests__/config.js`.

## Why
`configure()` is documented to accept a `Partial<Config>` and should merge
into existing config rather than reset omitted fields. The current
behavior makes RTL options order-dependent with DTL options, which is
surprising and a real footgun for users who set `reactStrictMode: true`
once at setup and later tweak DTL options per-test.

## Testing
Added a unit test that sets `reactStrictMode: true`, then configures only
`testIdAttribute`, and asserts both values are preserved. Did not run the
suite locally — `node_modules` is not installed in the sandbox. The
change is small and the new test fails before the fix and passes after by
inspection.

## Risk
Low — the diff is three lines of behavior change guarded by an
`!== undefined` check, plus one new test. No public API change.


Ziiyodullayevv (commented):
Good fix for strict mode preservation. Maintaining reactStrictMode when only DTL options are configured ensures consistent behavior.

PR #1463, opened 2026-05-18
feat: add renderAsync to fix React-Aria useLayoutEffect flushing (#1339)

### LINKED ISSUE #1339: React-Aria Component Combobox is not Fully rendering in Consuming Application Tests
### Our Various Versions of Things

- `@testing-library/react` version:
- Testing Framework and version:
		"@testing-library/jest-dom": "^5.16.5",
		"@testing-library/react": "^14.0.0",
		"@testing-library/user-event": "^14.4.3",
		"vitest": "^1.5.3"

- DOM Environment:
		"jsdom": "^21.1.0",

- React & Node Versions: 
		"@types/node": "18.14.6",
		"@types/react": "^18.2.45",
		"@types/react-dom": "18.0.11",

### Relevant code or config:

```Ts
it('should call to "MYAPI" when I enter values into my input', async () => {
		const user = userEvent.setup();
		const searchTerm = 'racc';
		const myContext = {...mockContext};
		renderContext(myContext)

		// click radio button for input to appear
		const showInputRadio = await screen.findAllByRole('radio', { name: 'Yes' });
		showInputRadio[0].click();
		// type our search in here
		// Here's the ISSUE spot: the input here is not being fully rendered / doesn't have all of the aria values & roles
		const input = screen.getByPlaceholderText('Type to Start Searching...');
		input.click();
		await user.type(input, searchTerm);

		await waitFor(() => {
			expect(fetchMock).toHaveBeenCalledWith(
				`/api/MYAPI?term=${searchTerm}`,
				buildRequestBody({}, 'GET')
			)
			expect(fetchMock).toHaveBeenCalledTimes(2);
		})
	})
```

### What you did:

We built a component library for internal use at our company using React-Aria components. We ingest / consume the library in a separate application.

### What happened / Problem Description:

Our components will render fully within the component library and can be tested to their full functionality: clicking, typing, searching, hit's api... etc. We confirmed that it works on our local & dev branches when running normally.

However, when we run our tests in the consuming application the input box that get's rendered in the testing dom is not fully filled out and the interactions do not properly execute. In this particular instance, we cannot type into the input to trigger the API call.

Example of Input rendered within our component tests...
<input
            aria-autocomplete="list"
            aria-controls="react-aria-:r3:"
            aria-expanded="true"
            aria-labelledby="react-aria-:r5:"
            autocomplete="off"
            autocorrect="off"
            class="w-full rounded-lg p-3 outline-none "
            data-focused="true"
            data-hovered="true"
            data-rac=""
            id="react-aria-:r4:"
            placeholder="Type to Start Searching..."
            role="combobox"
            spellcheck="false"
            title=""
            type="text"
            value="TestValue"
          />

Example of Input rendered within our consuming application using the same component...
<input
             class="w-full rounded-lg p-3 outline-none "
             data-rac=""
             placeholder="Type to Start Searching..."
             />

### Reproduction:

Unfortunately it is not possible in this situation to setup an example.

### Suggested solution, sorry, we don't really have one

We have tried a mix of userEvent.type and fireEvent.change to alter the input's value.
we have tried firing those events at different levels of the component.
We have tried selecting using different methodologies (find, query, get) on different values (test-id, placeholder, role... etc).

Ultimately, we are confused and do not know why it is not rendering fully in the tests even though it works within the application itself. It's a definite turning point for us because if we can't test the components that we create outside of the library we're gonna have to scrap it and start over.

Screenshot 2024-07-01 at 12 11 40 PM
Screenshot 2024-07-01 at 12 13 12 PM

Ayoub-glitsh:
## Problem

Closes #1339

Components like React-Aria `ComboBox` set ARIA attributes (`role`, `aria-expanded`, `aria-label`, etc.) inside `useLayoutEffect`, and often trigger a `setState` there — causing a second render cycle. The synchronous `render()` wraps the initial paint in `act()` but does **not** guarantee that the `useLayoutEffect → setState → re-render` chain completes before returning.

This makes the DOM appear half-initialised in consuming-application tests:

- `getByRole('combobox')` throws because `role` hasn't been set yet
- `aria-*` attributes are missing
- `userEvent.type()` has no effect because the input isn't fully wired up

The component works correctly at runtime and in the component library's own tests, but breaks in any consuming app's test suite.

## Solution

Add a new `renderAsync` utility that uses `await act(async () => {...})` to fully drain the microtask queue and all pending `useLayoutEffect` cycles before resolving.

### Usage

```js
// Before (broken with React-Aria):
const { getByPlaceholderText } = render(<MyComboBox />)
const input = getByPlaceholderText('Type to Start Searching...')
// role / aria-* attributes not set yet ❌

// After (fixed):
const { getByRole } = await renderAsync(<MyComboBox />)
const input = getByRole('combobox') // fully initialised ✓
await userEvent.type(input, 'racc')
await waitFor(() => expect(fetchMock).toHaveBeenCalledWith('/api/MYAPI?term=racc', ...))

### DIFF
--- a/src/__tests__/render-async.js
+++ b/src/__tests__/render-async.js
@@ -8,7 +8,8 @@
  * @see https://github.com/testing-library/react-testing-library/issues/1339
  */
 import * as React from 'react'
-import {renderAsync, screen, configure} from '../'
+import ReactDOMServer from 'react-dom/server'
+import {renderAsync, fireEvent, screen, configure} from '../'
 
 // ---------------------------------------------------------------------------
 // Helper: simulates a React-Aria-style component that sets ARIA attributes
@@ -129,6 +130,29 @@ describe('renderAsync', () => {
     await renderAsync(<Component />, {reactStrictMode: true})
     // In StrictMode effects run twice
     expect(effectSpy).toHaveBeenCalledTimes(2)
+  })
+
+  test('hydrate makes the UI interactive', async () => {
+    function App() {
+      const [clicked, handleClick] = React.useReducer(n => n + 1, 0)
+      return (
+        <button type="button" onClick={handleClick}>
+          clicked:{clicked}
+        </button>
+      )
+    }
+    const ui = <App />
+    const container = document.createElement('div')
+    document.body.appendChild(container)
+    container.innerHTML = ReactDOMServer.renderToString(ui)
+
+    expect(container).toHaveTextContent('clicked:0')
+
+    await renderAsync(ui, {container, hydrate: true})
+
+    fireEvent.click(container.querySelector('button'))
+
+    expect(container).toHaveTextContent('clicked:1')
   })
 
   test('legacyRoot throws when React does not support it', async () => {
--- a/src/__tests__/render-async.js
+++ b/src/__tests__/render-async.js
@@ -1,0 +1,171 @@
+/**
+ * Tests for renderAsync — the async variant of render that fully flushes
+ * useLayoutEffect chains before returning. This is the fix for components
+ * (e.g. React-Aria) that set ARIA attributes / roles in useLayoutEffect and
+ * therefore appear "incomplete" when queried immediately after a synchronous
+ * render() call in a consuming application's test suite.
+ *
+ * @see https://github.com/testing-library/react-testing-library/issues/1339
+ */
+import * as React from 'react'
+import {renderAsync, screen, configure} from '../'
+
+// ---------------------------------------------------------------------------
+// Helper: simulates a React-Aria-style component that sets ARIA attributes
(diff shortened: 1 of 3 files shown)


Ziiyodullayevv (commented):
Excellent addition. renderAsync fixes React-Aria useLayoutEffect flushing issues, improving compatibility with popular UI libraries.

eps1lon (changes requested):
The added test should work without any additional API. `act` already flushes state updates from inside Effects.

mdjastrzebski:
Afaik, real fix would be merging #1214. Adding renderAsync has been discussed [here](https://github.com/testing-library/react-testing-library/pull/1214#issuecomment-3490870546).

PR #1468, opened 2026-08-24
fix: Avoid act() re-entrant when dispatching events

### LINKED ISSUE #1231: 'event was not wrapped in act' warning with React 18, react-datepicker and user.type
### Relevant code or config:

```typescript
describe('ReactDatepicker', () => {
  beforeEach(() => (user = userEvent.setup()));

  it('should not emit act warnings', async () => {
    const now = new Date();
    const endOfNextYear = new Date(now.getFullYear() + 1, 11, 31);
    const dateFormat = 'M/d/yyyy';
    const userFormattedDate = format(endOfNextYear, dateFormat);
    const defaultFormattedDate = format(endOfNextYear, defaultDateFormat);

    render(<ReactDatepicker />);
    const input = screen.getByRole('textbox');
    await user.type(input, `${userFormattedDate}{Enter}`);
    expect(await screen.findByDisplayValue(defaultFormattedDate)).toBeInTheDocument()
  });

  it('should pass with fireEvent', async () => {
    const now = new Date();
    const endOfNextYear = new Date(now.getFullYear() + 1, 11, 31);
    const dateFormat = 'M/d/yyyy';
    const userFormattedDate = format(endOfNextYear, dateFormat);
    const defaultFormattedDate = format(endOfNextYear, defaultDateFormat);

    render(<ReactDatepicker />);
    const input = screen.getByRole('textbox');
    fireEvent.change(input, { target: { value: userFormattedDate } });
    fireEvent.keyDown(input, { key: 'Enter' });
    expect(await screen.findByDisplayValue(defaultFormattedDate)).toBeInTheDocument()
  });
});
```

### What you did:

use userEvent.type when targeting react-datepicker to enter a date. The weird thing is the "not wrapped in act" error only appears when a valid date is typed and the `{Enter}` key is pressed. This is apparency causing some special condition to happen within the implementation of react-datepicker which userEvent isn't act wrapping appropriately.

### What happened:

```
  console.error
    Warning: An update to r inside a test was not wrapped in act(...).

    When testing, code that causes React state updates should be wrapped into act(...):

    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */

    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
        at r (C:\git\pzaczkiewicz\forge-contrib-react-datepicker-user-event-test-failure\.yarn\__virtual__\react-datepicker-virtual-69b8f1e761\0\cache\react-datepicker-npm-4.16.0-c94778dc5d-59da0305ca.zip\node_modules\react-datepicker\dist\index.js:1:78951)
        at ReactDatepicker (C:\git\pzaczkiewicz\forge-contrib-react-datepicker-user-event-test-failure\packages\react-datepicker-user-event-test-failure\src\ReactDatepicker\ReactDatepicker.tsx:7:45)
```

### Reproduction:

https://github.com/pzaczkiewicz-athenahealth/react-datepicker-user-event-test-failure

### Problem description:

Not wrapped in act warnings are usually good indications that what you are testing is not what the user sees. Ignoring them may hide bigger issues in tests.

### Suggested solution:

Unknown. I dug into testing-library source code and couldn't completely wrap my head around why asyncWrapper and eventWrapper flip-flopped IS_REACT_ACT_ENVIRONMENT.

snowystinger:
**What**:

Closes https://github.com/testing-library/react-testing-library/issues/1231
And removes the need for https://github.com/testing-library/user-event/pull/1283

**Why**:

https://github.com/testing-library/user-event/pull/1283
> In React, nested synchronous act calls can prematurely teardown the act queue creating console warnings.

**How**:

By tracking if we're already inside an `act`, we can avoid wrapping it in a new one. At the same time, we can keep the wrapper around all events, which is an expectation using testing library's `configure`

**Checklist**:

- [ ] Documentation added to the N/A
- [x] Tests
- [x] Ready to be merged
      

Originally I was going to make this change over in user event, we have two PRs to do it, however, it's really a React and Preact concern. Changing it in User Event would affect everyone. Angular testing library already wraps all events for a different reason and so removing an event from that callback wrapper could be unexpected or breaking.  I would guess this change is also needed in the Preact testing library repo as well.


src/__tests__/event-wrapper-nesting.js
eps1lon, line 1: Can you write a test showing the underlying issue instead of testing implementation details (nesting depth)?

filipems:
There is now a React-only public-API reproduction of the underlying failure this PR is working around: https://github.com/react/react/issues/37374

The minimized case removes Testing Library, user-event, Radix, and focus/blur behavior entirely. A re-entrant `act()` entered while an outer `flushActQueue` is running sees `actScopeDepth === 0`, takes the outermost-exit path, cannot flush because `isFlushing` is already true, and nulls `actQueue` mid-flush. The outer `act()` then returns with React updates still pending on the real scheduler.

That issue includes a standalone `node repro.mjs`, expected vs actual behavior, causal controls, and the queue trace. It may also provide the behavioral test case requested in this PR without relying on act-depth instrumentation.

src/__tests__/event-wrapper-nesting.js
snowystinger: sorry about that, I was stumped on it for a little bit
I got there today though, just needed to sleep on it. done.

PR #1469, opened 2026-09-01
fix: preserve generic query element types in render result

korkt-kim:
**What**:

Preserve generic element types for the built-in queries returned by `render` and `pure.render`.

**Why**:

`RenderResult` currently wraps each query with `BoundFunction<Q[P]>`, which loses the generic signatures provided by DOM Testing Library's `BoundFunctions<Q>`.

As a result, calls such as the following do not type-check as expected:

```ts
view.getByRole<HTMLButtonElement>('button')
```

**How**:

Use `BoundFunctions<Q>` as the source for each mapped query property.

I kept the mapped property shape instead of intersecting with `BoundFunctions<Q>` directly because the direct intersection exposes built-in queries as methods and triggers `@typescript-eslint/unbound-method` for existing destructured queries.

Type tests were added for single- and multiple-element query variants.

**Checklist**:

      [docs site](https://github.com/testing-library/testing-library-docs) — N/A
- [x] Tests
- [x] Ready to be merged


korkt-kim:
Hi @eps1lon, could you take a look at this type-only fix when you have a chance?

It preserves generic element types for render-bound queries, with regression type tests. The validation matrix and CodeSandbox checks are passing; `codecov/project` is still waiting for a status.

The mapped-property shape is retained to avoid `unbound-method` lint regressions. Happy to adjust the approach or add coverage if needed. Thanks!

